PR #709, opened 2016-08-18
Remove encoding service

mangesh:
Removed entire encoding service ( encoding classes and its related functions ) from core rtMedia plugin.


app/admin/RTMediaAdmin.php
```diff
@@ -49,34 +49,25 @@ public function __construct() {
...
-			}
-
-			add_action( 'wp_ajax_rt_media_regeneration', array( $this, 'rt_media_regeneration' ), 1 );
-
 			if ( ! isset( $rtmedia->options ) ) {
 				$rtmedia->options = rtmedia_get_site_option( 'rtmedia-options' );
 			}
 
+			// Show admin notice to install transcoder plugin.
+			if ( ! class_exists( 'RT_Transcoder_Handler' ) ) {
```
chandra-patel: @mangesh Use `RT_Transcoder_Admin` instead of `RT_Transcoder_Admin`.

PR #754, opened 2016-10-20
After conflict remove

Gdad:
This is the code after conflict remove .
for multiple activity update and added filter for media menu tabs.


mangesh:
@Gdad Be careful next time when you send the PR. Make sure you are sending the PR on `develop` branch and not on `master`

PR #761, opened 2016-11-03
Translation issue - "Posted an update"

Gdad:
@mangesh  please check and merge this code.


mangesh:
@Gdad Please move this action in `rtmedia-actions.php` file
And the comment/doc block for the same function

Gdad:
@mangesh  please check code now .

PR #790, opened 2016-11-25
Permalink issue with plugin rtMedia activation and update

BhargavBhandari90:
Before page was showing page not found [404] issue after activating the plugin. We have to re-save permalink settings after activating plugin. 

Now no need to re-save the permalink setting.


index.php
```diff
@@ -113,6 +113,15 @@ function rtmedia_autoloader( $class_name ) {
 function is_rtmedia_vip_plugin() {
 	return ( defined( 'WPCOM_IS_VIP_ENV' ) && WPCOM_IS_VIP_ENV );
 }
+
+/**
+ * Chnage status for permalink settings when plugin is deactivated.
```
rahul286: Chnage => Change

PR #818, opened 2017-02-15
Solve pagination issue on likes, others and decuments tabs

BhargavBhandari90:


app/main/controllers/template/rtmedia-functions.php
```diff
@@ -2385,9 +2385,10 @@ function rtmedia_edit_media_privacy_ui() {
 
 	$privacymodel = new RTMediaPrivacy( false );
 	$privacy      = $privacymodel->select_privacy_ui( $echo = false );
-
-	if ( $privacy ) {
-		return "<div class='rtmedia-edit-privacy rtm-field-wrap'><label for='privacy'>" . esc_html__( 'Privacy : ', 'buddypress-media' ) . '</label>' . $privacy . '</div>';
+	if ( isset( $rtmedia_query->media['0']->privacy ) && $rtmedia_query->media['0']->privacy != '80' ) {
```
mangesh: @yahilmadakiya It is not necessary that `media['0']` will be set it will be an array, So accessing `media['0']` without checking it's existence will through fatal error.

app/main/controllers/template/rtmedia-functions.php
```diff
@@ -1515,10 +1507,20 @@ function rtmedia_pagination_page_link( $page_no = '' ) {
...
-	if ( isset( $rtmedia_query->action_query->media_type ) ) {
+	$allowed_types = array();
+
+	// get all allowed media type
+	foreach ( $rtmedia->allowed_types as $type ) {
+		$name = strtoupper( $type['name'] );
+		$allowed_types[] = constant( 'RTMEDIA_' . $name . '_SLUG' );
+	}
+
+	$media_type = apply_filters( 'rtmedia_media_type', $rtmedia_query->action_query->media_type );
```
mangesh: @BhargavBhandari90 It's not necessary that `media_type` will be always set.
Before accessing any object member make sure it exists.

BhargavBhandari90: This is fixed.

Issue #846, opened 2017-03-31
Comment in media option not Working for Mobile deveice


raftaar1191:
In mobile, Comment in media option is not working. 
When user open the media in popup and the media type is video or audio the uploder button in comment is not working.
![screenshot_2017-03-31-12-34-03](https://cloud.githubusercontent.com/assets/22215595/24539922/6d735d28-160e-11e7-91ba-8fba25111007.png)

oliziplok:
Do you have any updates on this?

PR #854, opened 2017-04-14
GH #846 Attach media button in comment not working in Safari in IOS mobile devices

rahul3883:
The problem is in Plupload library, plupload's input button with attribute type equals to file which is responsible for file selector to open is not clicked by plupload in safari in IOS mobile devices. So, by clicking it manually, file selector can be opened.


BhargavBhandari90:
This has been tested on ipad, ipod and anroid decides. And it is working fine.

rahul3883:
This has been tested and working fine.

Issue #859, opened 2017-04-28
Check edit media functionality when BuddyPress Activity Streams setting is disabled 


pranali333:
Whenever the user edits the media, gets `HTTP ERROR 500`.

Turning the BuddyPress Activity Streams setting ON solves the issue.

Please, look into the issue and debug the cause.

Thanks,

BhargavBhandari90:
This has been tested and working fine now.

rahul3883:
Fixed [here](https://github.com/BhargavBhandari90/rtMedia/commit/9dcf92f135a8a00cfd5813e9e7e8defaf981572c)

PR #860, opened 2017-04-28
Issue/GH #859

BhargavBhandari90:
Getting fatal error while editing media, if activity stream is disabled


BhargavBhandari90:
Tested and working fine now.

PR #908, opened 2017-06-02
GH#862 Show like button always

Sidsector9:
Shows like button whether comment section is enabled or not.

### DIFF
--- a/app/main/controllers/media/RTMediaLike.php
+++ b/app/main/controllers/media/RTMediaLike.php
@@ -67,6 +67,10 @@ class RTMediaLike extends RTMediaUserInteraction {
 		}
 	}
 
+	/**
+	 * This function displays the like button even if comment
+	 * section is disabled.
+	 */
 	function like_button_no_comments_callback() {
 		if ( empty( $this->media ) ) {
 			$this->init();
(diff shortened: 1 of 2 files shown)


app/main/controllers/media/RTMediaLike.php
```diff
@@ -66,6 +67,17 @@ function like_button_filter() {
 		}
 	}
 
+	function like_button_no_comments_callback() {
```
milindmore22: please add proper docblock / comment for the function.

templates/media/media-single.php
```diff
@@ -105,6 +105,17 @@ class="rtmedia-media-name <?php if ( rtmedia_album_name() ) { ?>rtmedia-media-na
...
+								<div class="rtmedia-actions-before-comments clearfix">
+									<div class="like-button-no-comments">
+										<?php do_action( 'like_button_no_comments' ); ?>
+									</div>
+								</div>
+								<div class="rtm-like-comments-info">
+									<?php show_rtmedia_like_counts(); ?>
+								</div>
+							</div>
 						<?php } ?>
```
milindmore22: this might be an extra space please check and remove it

milindmore22:
@chandrapatel please do a final review!

PR #933, opened 2017-07-07
AC-784 Add check if object to avoid notice error

Sidsector9:


app/main/controllers/template/rtmedia-functions.php
```diff
@@ -450,7 +450,9 @@ function rtmedia_type( $id = false ) {
 	} else {
 		global $rtmedia_media;
 
-		return $rtmedia_media->media_type;
+		if ( is_object( $rtmedia_media ) ) {
+			return $rtmedia_media->media_type;
+		}
```
kiranpotphode: @Sidsector9  As per doc block comment, function definition return int. In case of `if ( is_object( $rtmedia_media ) )` return false then function should return proper output.
Test code with new suggestion and update docblock comment for the function if necessory.

PR #939, opened 2017-07-13
Issue-GL-186-9 : Fix pagination in album search functionality.

kiranpotphode:
Remove redundant code in `rtMedia.backbone.js` and remove unnecessory condition.
Remove all filters used for search functionality to avoid modifying the query for current media results while getting the thumbnail of single media.

### DIFF
--- a/app/assets/js/rtMedia.backbone.js
+++ b/app/assets/js/rtMedia.backbone.js
@@ -400,7 +400,7 @@ jQuery( function( $ ) {
 			}
 
 			change_rtBrowserAddressUrl( href, '' );
-			galleryObj.getNext( nextpage, $( this ).parents( '#item-body' ), $( this ).parents( '.rtm-pagination' ) );
+			galleryObj.getNext( nextpage, $( this ).closest( '.rtmedia-container' ).parent(), $( this ).closest( '.rtm-pagination' ) );
 		} );
 
 		$( document ).on( 'submit', 'form#media_search_form', function( e ) {
--- a/app/assets/js/rtMedia.backbone.js
+++ b/app/assets/js/rtMedia.backbone.js
@@ -386,11 +386,6 @@ jQuery( function( $ ) {
 				page_base_url = $( this ).data( 'page-base-url' );
 				href = page_base_url + nextpage;
 				}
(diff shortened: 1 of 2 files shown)


app/main/controllers/template/RTMediaTemplate.php
```diff
@@ -219,6 +219,12 @@ function json_output() {
 		$media_array = array();
 
 		if ( $rtmedia_query->media ) {
+
+			// Remove all filters used for search functionality to avoid modifying query for current media results.
+			remove_filter( 'rtmedia-model-where-query', 'rtmedia_search_fillter_where_query', 10, 3 );
+			remove_filter( 'rtmedia-model-join-query', 'rtmedia_search_fillter_join_query', 11, 2 );
+			remove_filter( 'rtmedia-model-query-columns', 'rtmedia_model_query_columns', 10, 1 );
+
```
milindmore22: @kiranpotphode I think we should do add_filter for all three filters after foreach loop is completed.

app/assets/js/rtMedia.backbone.js
```diff
@@ -405,11 +400,7 @@ jQuery( function( $ ) {
 			}
 
 			change_rtBrowserAddressUrl( href, '' );
-			if ( $( this ).data( 'page-type' ) == 'num' ) {
-				galleryObj.getNext( nextpage, $( this ).parent().parent().parent().parent().parent(), $( this ).parent().parent() );
-			} else {
-				galleryObj.getNext( nextpage, $( this ).parent().parent().parent().parent().parent(), $( this ).parent().parent() );
-			}
+			galleryObj.getNext( nextpage, $( this ).parents( '#item-body' ), $( this ).parents( '.rtm-pagination' ) );
```
milindmore22: if possible can we use `closest` instead of `parents` it has better performance.
https://jsperf.com/jquery-parents-vs-closest/2

app/main/controllers/template/RTMediaTemplate.php
kiranpotphode: @milindmore22 Fixed

milindmore22:
@chandrapatel please do a final review!

PR #943, opened 2017-07-18
Improve localization strings

xavivars:
Hi,

It'd be very benefitial if strings were not sliced into pieces in order to add HTML tags in between, like

https://github.com/rtMediaWP/rtMedia/blob/develop/app/importers/RTMediaMigration.php#L381

This makes create good translations almost impossible.

Also, giving context to what the replacement strings do helps a lot.


app/importers/RTMediaMigration.php
```diff
@@ -378,11 +378,15 @@ function test() {
...
-				<p> <?php esc_html_e( 'Please Backup your ', 'buddypress-media' ); ?>
-					<strong><?php esc_html_e( 'DATABASE', 'buddypress-media' ); ?></strong>
-					<?php esc_html_e( ' and ', 'buddypress-media' ); ?>
-					<strong><?php esc_html_e( 'UPLOAD', 'buddypress-media' ); ?></strong>
-					<?php esc_html_e( ' folder before Migration.', 'buddypress-media' ); ?>
+				<p> 
+					<?php 
+			/* translators: %1$s gets replaced by '<strong>', and %2$s by '</strong>' */
+			echo sprintf(
+				__('Please Backup your %1$sDATABASE%2$s and %1$sUPLOAD%2$s folder before Migration.', 'buddypress-media' )
```
xavivars: Sorry about hte Syntax error. Regarding the WP Coding Standars, I already tried to follow them (i.e. https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#opening-and-closing-php-tags) but probably missed something.

app/importers/RTMediaMigration.php
```diff
@@ -378,11 +378,14 @@ function test() {
...
 			<div class="error">
-				<p> <?php esc_html_e( 'Please Backup your ', 'buddypress-media' ); ?>
-					<strong><?php esc_html_e( 'DATABASE', 'buddypress-media' ); ?></strong>
-					<?php esc_html_e( ' and ', 'buddypress-media' ); ?>
-					<strong><?php esc_html_e( 'UPLOAD', 'buddypress-media' ); ?></strong>
-					<?php esc_html_e( ' folder before Migration.', 'buddypress-media' ); ?>
+				<p> 
+					<?php 
+			/* translators: %1$s gets replaced by '<strong>', and %2$s by '</strong>' */
+			echo sprintf(
```
pradeep910: @BhargavBhandari90 Fix the indentation.

BhargavBhandari90:
@pradeep910 requested change has been done now. Please check now.

xavivars:
This is a single message, but there are plenty of them in the plugin, so (ideally) IMHO all messages should be changed to follow this pattern instead.

pradeep910:
@xavivars True. Thanks for pointing it out. 

@BhargavBhandari90 Please do the same thing at other places, where we have multiple translation function calls.

BhargavBhandari90:
Will continue in https://github.com/rtMediaWP/rtMedia/pull/990

Issue #948, opened 2017-07-24
`media_title="true"` not working for shortcodes


BhargavBhandari90:
When we add `media_title="true"` to the shortcode, then it should show title for media. But it is not displaying the title.

BhargavBhandari90:
PR for this: https://github.com/rtMediaWP/rtMedia/pull/956

pranali333:
Fixed and released with version 4.4.3

PR #956, opened 2017-07-27
Fix - issue #948

BhargavBhandari90:
media_title attribute was not working for shortcodes

It was not showing title even it's value set to true

### DIFF
--- a/app/main/routers/query/RTMediaQuery.php
+++ b/app/main/routers/query/RTMediaQuery.php
@@ -602,6 +602,10 @@ class RTMediaQuery {
 	}
 
 	function populate_media() {
+		global $rtmedia_query;
+
+		// Check if the page is gallery shortcode or not.
+		$is_gallery_shortcode = ( isset( $rtmedia_query->is_gallery_shortcode ) && true === $rtmedia_query->is_gallery_shortcode ) ? true : false;
 
 		$this->set_privacy();
 		if ( $this->is_single() ) {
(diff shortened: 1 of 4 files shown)


app/main/routers/query/RTMediaQuery.php
```diff
@@ -676,13 +680,13 @@ function populate_media() {
...
 		}
 
-		// media title option
-		if ( isset( $this->media_query['media_title'] ) ) {
-			if ( 'false' === $this->media_query['media_title'] ) {
-				// Add filter show media title
+		// Media title option.
+		if ( isset( $this->media_query ) ) {
+			if ( ( isset( $this->media_query['media_title'] ) && 'false' === $this->media_query['media_title'] ) || ( true === $is_gallery_shortcode && ! isset( $this->media_query['media_title'] ) ) ) {
+				// Add filter show media title.
```
kiranpotphode: Just a suggestion you can add dock block for hooks as per this reference
https://github.com/buddypress/BuddyPress/blob/master/src/bp-core/bp-core-attachments.php#L105
@BhargavBhandari90

app/main/routers/query/RTMediaQuery.php
```diff
@@ -676,13 +680,13 @@ function populate_media() {
 			unset( $this->media_query['lightbox'] );
 		}
 
-		// media title option
-		if ( isset( $this->media_query['media_title'] ) ) {
-			if ( 'false' === $this->media_query['media_title'] ) {
-				// Add filter show media title
+		// Media title option.
+		if ( isset( $this->media_query ) ) {
+			if ( ( isset( $this->media_query['media_title'] ) && 'false' === $this->media_query['media_title'] ) || ( true === $is_gallery_shortcode && ! isset( $this->media_query['media_title'] ) ) ) {
```
pradeep910: @BhargavBhandari90 `'false' === $this->media_query['media_title']`
I think `false` should be without single quotes. 
Please check again if its boolean or not.

BhargavBhandari90: This is a shortcode attribute value which will be a string.

pradeep910: Ok thanks for clarification.

Issue #941, opened 2017-07-18
Upload just falls over while using special characters into filename


BhargavBhandari90:
A problem uploading media via media uploader which contains the ^ character in the filename. 
No issue with backend media upload. 
Test by uploading a file named "пёсики-песочница^610680".

BhargavBhandari90:
Not solved in activity.

pranali333:
Fixed and released with version 4.4.3

PR #959, opened 2017-08-02
Issue/GH #941

BhargavBhandari90:
Image was not showing in which special characters are used.


app/main/controllers/activity/RTMediaActivity.php
```diff
@@ -122,7 +122,7 @@ function media( $media, $type = 'activity' ) {
 				$thumbnail_id = $media->media_id;
 				if ( $thumbnail_id ) {
 					list( $src, $width, $height ) = wp_get_attachment_image_src( $thumbnail_id, apply_filters( 'rtmedia_activity_image_size', 'rt_media_activity_image' ) );
-					$html = '<img alt="' . esc_attr( $media->media_title ) . '" src="' . esc_url( $src ) . '" />';
+					$html = '<img alt="' . esc_attr( $media->media_title ) . '" src="' . set_url_scheme( $src ) . '" />';
```
pradeep910: @BhargavBhandari90 I think set url scheme will just set the scheme (http, https). So you still need esc_url.
See this - https://developer.wordpress.org/reference/functions/set_url_scheme/

BhargavBhandari90: @pradeep910 

That code is added by checking the code from "wp-admin/includes/media.php:2740".
Because in media library, it was not breaking.

We should use `esc_url`. But WP is inserting name with special character.So it is the issue from WP.

The issue is reported here: https://core.trac.wordpress.org/ticket/15955 and https://core.trac.wordpress.org/ticket/41353

BhargavBhandari90: We can use `esc_url` after solution of https://core.trac.wordpress.org/ticket/15955

BhargavBhandari90:
@pradeep910 

That code is added by checking the code from "wp-admin/includes/media.php:2740".
Because in media library, it was not breaking.

We should use esc_url. But WP is inserting name with special character.So it is the issue from WP.

The issue is reported here: https://core.trac.wordpress.org/ticket/15955 and https://core.trac.wordpress.org/ticket/41353

We can use esc_url after solution of https://core.trac.wordpress.org/ticket/15955

BhargavBhandari90:
@pradeep910 Please check above comment and let me know if any doubt.

pradeep910:
Thanks for the info @BhargavBhandari90

PR #975, opened 2017-09-11
Master

sumeetsarna:
Update .gitignore file

Update Browserstack badge key

Update Travis file

Merge codeception code from develop branch


tests/codeception/tests/_bootstrap.php
```diff
@@ -0,0 +1,8 @@
+<?php
+// This is global bootstrap for autoloading
```
juhi123: @sumeetsarna Remove commented code

tests/codeception/tests/_support/Page/BuddypressSettings.php
```diff
@@ -0,0 +1,232 @@
+<?php
+
+namespace Page;
+
+use Page\Constants as ConstantsPage;
+
+class BuddypressSettings {
+
+	public static $userProfileLink = 'a#user-xprofile';
```
juhi123: @sumeetsarna Why we are declaring selectors here, @kjnanda was following the approach of keeping the all selectors in constants.php

sumeetsarna: @juhi123 This is @kjnanda's code. 

And these are Page objects and that's how they are used http://codeception.com/docs/06-ReusingTestCode#PageObjects

juhi123: @sumeetsarna Yeah I just noticed, it's @kjnanda's code only. And I know it's page object and so is Constants.php. We were following this approach to have all selectors in one page objects that is Constants.php. Somehow she missed it, in the same file you will see selectors are being called from Constants.php

sumeetsarna: @juhi123 We can incorporate this suggestion in the next iteration. Doesn't seem to be a Blocker.  It will cause too many changes which could be taken up in next iteration.

juhi123: @sumeetsarna Agree it's not a blocker

tests/codeception/tests/_bootstrap.php
sumeetsarna: Thanks @juhi123.  I've made the changes. Please check.

BhargavBhandari90:
No need to merge this as already merged this code in previous PRs.

Issue #984, opened 2017-09-22
Issue with BuddyPress and rtMedia when we use the action `bp_before_member_header` for showing shortcode


BhargavBhandari90:
If we use the given code, then it creates the problem

```
function rtmc_show_gallary() {

    if ( ! function_exists( 'bp_displayed_user_id' ) && ( function_exists( 'bp_is_user' ) && ! bp_is_user() ) ) {
        return;
    }
    $user_id = bp_displayed_user_id();
    echo do_shortcode( '[rtmedia_gallery global="true"]' );

}
add_action( 'bp_before_member_header', 'rtmc_show_gallary' );

BhargavBhandari90:
@pranali333 https://github.com/rtMediaWP/library/blob/master/media/RTMC_Gallery_Shortcode.php

kjnanda:
@BhargavBhandari90 Please check my comment below.

1. If I used this snippet, I am able to see the Gallery media above the Cover image. But media comes in `verticle` alignment. Which doesn't look proper. 
Please check this link - https://goo.gl/yp2BrW

2. After I add code snippet, I am not able to see the `Upload panel` on the front end. Which should not happen.
Please check this link - https://goo.gl/VfqRVZ

BhargavBhandari90:
@kjnanda Ignore the code given in first comment. This has been fixed in https://github.com/rtMediaWP/library/blob/master/media/RTMC_Gallery_Shortcode.php.

BhargavBhandari90:
> If I used this snippet, I am able to see the Gallery media above the Cover image. But media comes in verticle alignment. Which doesn't look proper.
Please check this link - https://goo.gl/yp2BrW

This is not our problem as this is custom code.

kjnanda:
@BhargavBhandari90 In reference of this - 

> @kjnanda Ignore the code given in the first comment. This has been fixed in https://github.com/rtMediaWP/library/blob/master/media/RTMC_Gallery_Shortcode.php.

I can see `Upload Panel` using this code.

Done. 👍

PR #985, opened 2017-09-26
GH#984 Added check for user group

Sidsector9:
See #984 
This only solves the fatal error. The missing media gallery due to shortcode will be fixed in the next update.

### DIFF
--- a/app/main/controllers/template/RTMediaNav.php
+++ b/app/main/controllers/template/RTMediaNav.php
@@ -188,7 +188,12 @@ class RTMediaNav {
 
 	public function sub_nav() {
 		global $rtmedia, $rtmedia_query;
-		if ( function_exists( 'bp_is_group' ) && bp_is_group() ) {
+
+		$active_components = bp_get_option( 'bp-active-components' );
+		$user_groups = $active_components['groups'];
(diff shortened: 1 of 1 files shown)


app/main/controllers/template/RTMediaNav.php
```diff
@@ -188,7 +188,12 @@ function admin_nav() {
 
 	public function sub_nav() {
 		global $rtmedia, $rtmedia_query;
-		if ( function_exists( 'bp_is_group' ) && bp_is_group() ) {
+
+		$active_components = bp_get_option( 'bp-active-components' );
+		$user_groups = $active_components['groups'];
+		$user_group_status = ( '1' === $user_groups ) ? true : false;
```
milindmore22: add proper alignment

app/main/controllers/template/RTMediaNav.php
```diff
@@ -265,23 +270,25 @@ public function sub_nav() {
...
 						$profile
 					)
 				);
 			} else {
-				$profile_link = trailingslashit(
-					get_rtmedia_group_link(
-						$profile
-					)
-				);
+				if ( $user_group_status ) {
```
milindmore22: can we use `elseif` instead

PR #987, opened 2017-10-03
Develop

sumeetsarna:
Update Travis file to fix WP CLI error

Update to use latest BuddyPress plugin.

Followed workaround mentioned here https://easyengine.io/blog/solution-couldnt-extract-wordpress-archive-error/


juhi123:
It looks good :)

PR #989, opened 2017-10-04
Fix WP-CLI Error and Add Testcase for Gallery Media Search

kjnanda:
1. Fix WP-CLI installation error
2. Remove duplicate code in travis file
3. Add test cases for Enable/Disable Gallery Media Search


juhi123:
Looks good 👍

PR #990, opened 2017-10-09
Xavivars patch 1

BhargavBhandari90:
For https://github.com/rtMediaWP/rtMedia/pull/943

### DIFF
--- a/app/importers/RTMediaMigration.php
+++ b/app/importers/RTMediaMigration.php
@@ -366,7 +366,7 @@ class RTMediaMigration {
 		global $rtmedia_error;
 		if ( isset( $rtmedia_error ) && true === $rtmedia_error ) {
 			?>
-			<div class="error"><p> Please Resolve create database error before migration.</p></div>
+			<div class="error"><p><?php esc_html__( 'Please Resolve create database error before migration.', 'buddypress-media' ); ?></p></div>
 			<?php
 		}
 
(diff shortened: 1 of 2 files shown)


app/importers/RTMediaMigration.php
```diff
@@ -366,7 +366,7 @@ function test() {
 		global $rtmedia_error;
 		if ( isset( $rtmedia_error ) && true === $rtmedia_error ) {
 			?>
-			<div class="error"><p> Please Resolve create database error before migration.</p></div>
+			<div class="error"><p><?php esc_html__( 'Please Resolve create database error before migration.', 'buddypress-media' ); ?></p></div>
```
milindmore22: needs to be `esc_html_e` or `echo esc_html__`

app/importers/RTMediaMigration.php
```diff
@@ -378,11 +378,11 @@ function test() {
...
 			<div class="error">
-				<p> <?php esc_html_e( 'Please Backup your ', 'buddypress-media' ); ?>
-					<strong><?php esc_html_e( 'DATABASE', 'buddypress-media' ); ?></strong>
-					<?php esc_html_e( ' and ', 'buddypress-media' ); ?>
-					<strong><?php esc_html_e( 'UPLOAD', 'buddypress-media' ); ?></strong>
-					<?php esc_html_e( ' folder before Migration.', 'buddypress-media' ); ?>
+				<p>
+					<?php
+					/* translators: %1$s gets replaced by '<strong>', and %2$s by '</strong>' */
+					echo sprintf( esc_html__( 'Please Backup your %1$sDATABASE%2$s and %1$sUPLOAD%2$s folder before Migration.', 'buddypress-media' ), '<strong>', '</strong>' );
```
milindmore22: you can use `printf ` here to avoid excessive processing

BhargavBhandari90: @milindmore22 Can you share something about `excessive processing`?

milindmore22: in simple terms 
`printf` will directly output values 
`sprintf ` will return values and `echo` does output job

so by directly using `printf` to output values, we are saving excessive processing of returning values and printing them.

Issue #978, opened 2017-09-20
Styling needs improvement when masonry is disabled


sumeetsarna:
When Masonry grid layout is enabled Media page looks like:

![image](https://user-images.githubusercontent.com/8191145/30637127-1839981a-9e15-11e7-8dd6-736a2916b2ef.png)

But when it is disabled it looks like:

![image](https://user-images.githubusercontent.com/8191145/30637514-2386d9a2-9e16-11e7-9fcb-9356428e4b1b.png)

In this case above Media Size Settings were:

![image](https://user-images.githubusercontent.com/8191145/30637575-49e91222-9e16-11e7-91c4-38e1e51315aa.png)

If we use Media Size Settings same as https://github.com/rtMediaWP/rtMedia/issues/976

then when Masonry is disabled it looks like:

![image](https://user-images.githubusercontent.com/8191145/30638344-bfab13a0-9e18-11e7-9ada-babf50880a51.png)

![image](https://user-images.githubusercontent.com/8191145/30638413-0004bd0c-9e19-11e7-9555-7e4bd2fccfcd.png)

Please check **Edit and Delete** on mouseover and the count.

Checked on Twenty Seventeen theme.

kjnanda:
@BhargavBhandari90 Please check my comment below.

1. Check styling when `Masonry layout` is disabled. **`Done`**
Looks good to me 👍 

2. Check styling when media size width set to **`Zeor`** -->**UI still caters for this**. 

Please check below screenshots.

- UI for Albums - https://goo.gl/5dA7XT

- UI for All seperate Media -  https://goo.gl/vNufaM

BhargavBhandari90:
@kjnanda  This has been fixed

kjnanda:
@BhargavBhandari90 This looks good to me. 👍

sumeetsarna:
@BhargavBhandari90  It still needs improvement.

For this Media Size Settings mentioned above:

![image](https://user-images.githubusercontent.com/8191145/31482242-268880b6-af45-11e7-8f76-54dbfb02133e.png)

The All Media page looks like:

![image](https://user-images.githubusercontent.com/8191145/31482232-184cc714-af45-11e7-981d-fd3976dcf26f.png)

Please check:

* [ ] 1.Edit and Delete on mouseover. The width should be same as the image.

* [ ] 2. The hover color should be lighter. If you check the image above `Delete` icon and first letter isn't properly visible

* [ ] 3. Albums display needs improvement

[...]

PR #994, opened 2017-10-11
Issue/GH #978

BhargavBhandari90:
Fix for: https://github.com/rtMediaWP/rtMedia/issues/978

### DIFF
--- a/app/main/RTMedia.php
+++ b/app/main/RTMedia.php
@@ -312,11 +312,11 @@ class RTMedia {
 
 		// If height or width given zero, then show it's original height or with.
 		if ( '0' === $height ) {
-			$media_height = '100%';
+			$media_height = 'auto';
 		}
 
 		if ( '0' === $width ) {
-			$media_width = '100%';
+			$media_width = 'auto';
 		}
 
 		?>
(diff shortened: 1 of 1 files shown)


app/main/RTMedia.php
```diff
@@ -343,30 +343,30 @@ function custom_style_for_activity_image_size() {
...
+		$width  = ( isset( $this->options['defaultSizes_photo_thumbnail_width'] ) ) ? $this->options['defaultSizes_photo_thumbnail_width'] : '0';
 		// Get height from rtMedia settings.
-		$height = ( isset( $this->options['defaultSizes_photo_medium_height'] ) ) ? $this->options['defaultSizes_photo_medium_height'] : '0';
+		$height = ( isset( $this->options['defaultSizes_photo_thumbnail_height'] ) ) ? $this->options['defaultSizes_photo_thumbnail_height'] : '0';
 
 		$media_height = $height . 'px';
 		$media_width  = $width . 'px';
 
 		// If height or width given zero, then show it's original height or with.
 		if ( '0' === $height ) {
```
milindmore22: this condition never will be true
if there is no `defaultSizes_photo_thumbnail_height` set then it will take `0`
Line [351](https://github.com/rtMediaWP/rtMedia/pull/994/files#diff-f23e65e07ebeccfcd8de51efd0419621R350) makes it `0px`

same goes for width

BhargavBhandari90: @milindmore22 At line 351, I assigned a value to a different variable `$media_height` which will become `0px`. `$height` will remain `0` only.

Let me know if I misunderstood something.

milindmore22: ahh sorry, my bad

PR #1015, opened 2017-12-11
GH#1004 Fix style for back button

Sidsector9:


thrijith:
Fixed in https://github.com/rtMediaWP/rtMedia/pull/1023

PR #1016, opened 2017-12-11
GH#1007 Added min-height to avoid overflow:hidden blockage

Sidsector9:


thrijith:
Fixed in https://github.com/rtMediaWP/rtMedia/pull/1032

Issue #1019, opened 2017-12-13
Music media type catered UI 


kjnanda:
![screenshot-krupa rtdemo in 2017-12-13 12-24-18-735](https://user-images.githubusercontent.com/15173772/33925601-9b474fc2-e000-11e7-96aa-d508411d6eea.png)

![screenshot-krupa rtdemo in 2017-12-13 12-34-47-788](https://user-images.githubusercontent.com/15173772/33925920-1817f7e4-e002-11e7-9512-935e0a9b1fec.png)

![unnamed image 3](https://user-images.githubusercontent.com/15173772/33925922-1b71d126-e002-11e7-9b25-3bdca888baee.png)

![screenshot-krupa rtdemo in 2017-12-13 13-01-36-278](https://user-images.githubusercontent.com/15173772/33926901-0a379bf8-e006-11e7-988e-b95b84832c34.png)

**Test Environment**

1. Browser - Chrome
2. WP version - 4.9.1

kjnanda:
@thrijith Still UI breaks when media uploaded on comment. Only Media upload on `activity` UI seems good. Please check this visual for reference.

![unnamed image](https://user-images.githubusercontent.com/15173772/34095377-31131564-e3f7-11e7-9fba-2396cd0aab94.png)

![untitled gif 1](https://user-images.githubusercontent.com/15173772/34095443-72722054-e3f7-11e7-90cb-272eefcf43d7.gif)

pranali333:
Hi @SID177 ,

Please also check if this issue has been fixed/released.

Thanks

kjnanda:
Most of the issues mentioned in the subtask are fixed. I will recreate a separate issue for the things which are not covered yet.

PR #1022, opened 2017-12-15
 fix icon placement issue in #1019 for audio/video player

thrijith:
fixed icon placement issue in #1019 for audio/video player


thrijith:
fixed padding issues in rtMedia.js

thrijith:
don't merge the pull request right now it needs more changes
it only has partial fixes to the mentioned issue

PR #1043, opened 2018-01-17
upgrade mediaelement

thrijith:
updated to 4.2.8


app/main/RTMedia.php
```diff
@@ -990,14 +990,11 @@ function create_table_error_notice() {
 
 	function enqueue_scripts_styles() {
 		global $rtmedia;
```
thrijith: @sanketio  @PatelUtkarsh  please see if this is ok

app/main/RTMedia.php
```diff
@@ -990,14 +990,11 @@ function create_table_error_notice() {
...
-		if ( wp_script_is( 'wp-mediaelement', 'registered' ) ) {
-			wp_enqueue_style( 'wp-mediaelement' );
-			wp_enqueue_script( 'wp-mediaelement' );
-		} else {
-			wp_enqueue_script( 'wp-mediaelement', RTMEDIA_URL . 'lib/media-element/mediaelement-and-player.min.js', '', RTMEDIA_VERSION );
-			wp_enqueue_style( 'wp-mediaelement', RTMEDIA_URL . 'lib/media-element/mediaelementplayer.min.css', '', RTMEDIA_VERSION );
-			wp_enqueue_script( 'wp-mediaelement-start', RTMEDIA_URL . 'lib/media-element/wp-mediaelement.js', 'wp-mediaelement', RTMEDIA_VERSION, true );
-		}
+
+		wp_deregister_script('wp-mediaelement');
```
PatelUtkarsh: Instead of deregistering can we enqueue our own script?

PatelUtkarsh: `wp-mediaelement` use the different handle.

thrijith: when i tried without  deregistering core js  gets loaded which is causing issues

PatelUtkarsh: With different handle?

PR #1054, opened 2018-03-05
Issue/media single page comment issue 152

SID177:
- The issue is solved
- The error message is moved down to comment form
- When error is raised, the empty comment box will be focused automatically.


app/assets/js/rtMedia.js
```diff
@@ -989,7 +989,44 @@ function rtmedia_single_media_alert_message( msg, action ) {
 		action_class = 'rtmedia-warning';
 	}
 
-	jQuery( '.rtmedia-single-media .rtmedia-media' ).css( 'opacity', '0.2' );
+	/**
+	 * Fixed code
```
chandrapatel: Add a comment about which issue you've fixed instead of just `Fixed code` comment.

app/assets/js/rtMedia.js
```diff
@@ -989,7 +989,44 @@ function rtmedia_single_media_alert_message( msg, action ) {
...
+	 */
+	var comment_form = jQuery( '#rt_media_comment_form' );
+	var msg_container = jQuery( '.rtmedia-message-container' );
+	if( comment_form.next().attr('class') === 'rtmedia-message-container' ){
+		msg_container.remove();
+		comment_form.after( '<div title="Click to dismiss" class=\'rtmedia-message-container\' style="margin:1em 0;"><span style="border: 2px solid #884646;" class=\'' + action_class + '\'>' + msg + ' </span></div>' );
```
chandrapatel: Don't construct HTML strings. Instead of it, Create HTML elements using jQuery.

app/assets/js/rtMedia.js
```diff
@@ -989,7 +989,44 @@ function rtmedia_single_media_alert_message( msg, action ) {
...
+
+		msg_container = jQuery( '.rtmedia-message-container' );
+		var span = msg_container.find('span');
+		setTimeout(function(){
+			span.css({border:'none'});
+		},500);
+
+	} else {
+		comment_form.after( '<div title="Click to dismiss" class=\'rtmedia-message-container\' style="margin:1em 0;"><span class=\'' + action_class + '\'>' + msg + ' </span></div>' );
```
chandrapatel: Don't construct HTML strings. Instead of it, Create HTML elements using jQuery.

app/assets/js/rtMedia.js
```diff
@@ -989,7 +989,44 @@ function rtmedia_single_media_alert_message( msg, action ) {
...
+
+	msg_container.click( function() {
+		msg_container.remove();
+	} );
+
+
+	/**
+	 * Issued code
+	 */
+	/*jQuery( '.rtmedia-single-media .rtmedia-media' ).css( 'opacity', '0.2' );
```
chandrapatel: Don't keep old code. If it's not needed then remove it.

app/assets/js/rtMedia.js
SID177: Element created

app/assets/js/rtMedia.js
SID177: Element created

app/assets/js/rtMedia.js
SID177: Removed old code

app/assets/js/rtMedia.js
SID177: Comment fixed

Issue #1059, opened 2018-03-07
Negative comment count


SID177:
When deleting attached media comment from activity, the comment count becomes negative (false). 

**Steps to generate issue:**
1. Check `Enable media in comment` from `rtMedia -> Settings -> BuddyPress` section.
2. Add comment with media attachment on any activity and note the comment count
3. Delete the media comment, and cross check the comment count with number of comments
![screenshot-negative-comment](https://user-images.githubusercontent.com/23107546/37091353-d373ec7e-222d-11e8-8aab-37d1968ab42f.png)

juhi123:
Looks good now

PR #1060, opened 2018-03-07
Issue/negative comment count 1059

SID177:
Issue link: #1059

- The buddypress-comment-js is getting confused between media attachment ul and child comment ul in activity comment.
- Solution is to remove media attachment uls when the comment is being removed


app/assets/js/rtMedia.backbone.js
```diff
@@ -8,6 +8,31 @@ var rtmedia_load_template_flag = true;
...
+	 */
+	$( document ).ready( function () {
+		/**
+		 * Bind dynamic event on delete button to remove media ul
+		 */
+		$( '#activity-stream' ).on( 'click', '.acomment-delete', function () {
+			/**
+			 * get media ul
+			 */
+			let media_children = $( this ).parent().parent().find( 'div.acomment-content ul.rtmedia-list' );
```
milindmore22: instead of `parent().parent()` please use `closest();`
https://api.jquery.com/closest/

PR #1079, opened 2018-03-24
update code to fix warning

thrijith:
ref `GL-21` in rtmedia-moderation add-on.


app/main/controllers/template/rtmedia-functions.php
```diff
@@ -356,7 +356,11 @@ function rtmedia_media_id( $id = false ) {
 		$media = $model->get_media( array(
 			'id' => $id,
 		), 0, 1 );
-		return $media[0]->media_id;
+
+		// update code to fix GL-21-rtmoderation.
+		if ( isset( $media[0] ) && ! empty( $media[0] ) ) {
```
milindmore22: I think not empty check is enough here

PR #1080, opened 2018-03-26
Message after comment form only for empty comment

SID177:
- The empty comment message will be shown after comment box only if the comment is empty.
- Other messages will be shown to its regular place.

Issue: GL-211


abhijitrakas:
Looks good to me.

PR #1087, opened 2018-04-03
fix number count format

montu1996:


app/main/controllers/template/rtmedia-functions.php
```diff
@@ -4154,3 +4154,23 @@ function is_rtmedia_upload_other_enabled() {
...
+
+/**
+ * Format a number to human readable format.
+ *
+ * @param Integer $n Number to be formatted.
+ * @return String Return format a number to human readable format.
+ */
+function number_to_human_readable( $n ) {
+	if ( $n >= 1000000000000 ) {
+		return round( ( $n / 1000000000000 ), 1 ) . 'T';
```
thrijith: @chandrapatel  need suggestion for this, should this character be translatable ?

PR #1098, opened 2018-04-19
Gh 1096 extra get request sent to server

umeshnevase:


app/main/controllers/template/rtmedia-actions.php
```diff
@@ -790,22 +790,6 @@ function rt_check_addon_status() {
 				}
 			}
 
-			// Listen for activate button to be clicked
-
-			// Also check if information about the addon in already fetched from the store
-			// If it's already fetched, then don't send the request again for the information
-
-			if ( ! empty( $addon_active ) && ! isset( $_POST[ 'edd_' . $addon_id . '_license_activate' ] ) ) {
```
SID177: Use `filter_input( INPUT_POST, ... )` here, then check for `! empty( $varname )` @umeshnevase

umeshnevase:
This commit is reverted because it was not tested and may cause the other license verification functionality.

PR #1108, opened 2018-05-28
Fix Under the like tab in media, album list was not getting populted

malavasita:
Fixed Under the like tab in media, in uploader section albums does not get populated in album dropdown.

### DIFF
--- a/app/helper/RTMediaModel.php
+++ b/app/helper/RTMediaModel.php
@@ -41,6 +41,7 @@ class RTMediaModel extends RTDBModel {
 	 */
 	function get( $columns, $offset = false, $per_page = false, $order_by = 'media_id desc', $count_flag = false ) {
 		global $wpdb;
+		global $rtmedia_interaction;
 		$select = 'SELECT ';
 		if ( $count_flag ) {
 			$select .= 'count(*) ';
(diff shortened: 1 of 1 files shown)


app/helper/RTMediaModel.php
```diff
@@ -107,8 +108,11 @@ function get( $columns, $offset = false, $per_page = false, $order_by = 'media_i
 		}
 
 		$select    = apply_filters( 'rtmedia-model-select-query', $select, $this->table_name );
-		$join      = apply_filters( 'rtmedia-model-join-query', $join, $this->table_name );
-		$where     = apply_filters( 'rtmedia-model-where-query', $where, $this->table_name, $join );
+		$bp_template = get_option( '_bp_theme_package_id' );
+		if ( empty( $bp_template ) && 'nouveau' !== $bp_template && "likes" === $rtmedia_interaction->routes[ 'media' ]->query_vars[ 0 ] ) {
```
milindmore22: remove space from the array keys, use single quote

app/helper/RTMediaModel.php
```diff
@@ -107,8 +108,11 @@ function get( $columns, $offset = false, $per_page = false, $order_by = 'media_i
...
+			$join      = apply_filters( 'rtmedia-model-join-query', $join, $this->table_name );
```
milindmore22: align properly.

PR #1118, opened 2018-06-12
Fix issue with user deletion

thrijith:
fixes GL-217

Fix issue when deleting a user on a multi-site setup.


juhi123:
Tested, looks good to me :)

PR #1133, opened 2018-06-19
Fix upload media issue with activity conversation

malavvasita:
Fixed upload media in activity view conversation


app/assets/js/rtMedia.backbone.js
```diff
@@ -2646,8 +2646,8 @@ function rtmedia_comment_media_upload( upload_comment ){
 function rtmedia_activity_stream_comment_media(){
 
     // For Buddypress new template nouveau
```
kiranpotphode: Change comment as per code code changes.

malavvasita: Doc will be same. Just reverted condition.
Else this function will work as it is. @kiranpotphode

Issue #838, opened 2017-03-22
rtmedia_comments_enabled() Function may generate notice.


yahilmadakiya:
The **rtmedia_comments_enabled()** function may generate notice/fatal. This should be like **get_rtmedia_default_privacy()** function. 

Ref: https://github.com/rtMediaWP/rtMedia/blob/develop/app/main/controllers/template/rtmedia-functions.php#L2542

thrijith:
Fixed in rtMedia version 4.5.1

PR Ref: [#1139](https://github.com/rtMediaWP/rtMedia/pull/1139)

PR #1139, opened 2018-06-20
Add check for rtmedia_comments_enabled

thrijith:
Add a check in rtmedia_comments_enabled to handle a possible error situation.

ref #838


app/main/controllers/template/rtmedia-functions.php
```diff
@@ -1708,15 +1708,19 @@ function rtmedia_get_pagination_values() {
 /**
  * Checking if comments are enabled
```
kiranpotphode: More descriptive docblock will be appreciated. Correct as per phpcs standards

PR #1144, opened 2018-06-21
fix GL-222

montu1996:
Fix Media search


malavvasita:
This issue is duplicate of PR: #1154

Issue #223, opened 2013-03-05
Gallery Wall


webdesignerart:
First of all the plugin is very nice and professional and at first use i became a fan of it. Well i am seeing as far this plugin is for BuddyPress means community based site, many users upload images but there is no any seperate wall page or gallery page that have recent/most likes/category wise pictures sorting.

I think this will be very important part of this plugin based on community site.

Regards

PR #1149, opened 2018-06-25
Fixed undefined $ console error in upload terms

malavvasita:
Uncaught TypeError: $ is not a function at HTMLDocument while upload terms is enable in rtMedia.

Issue Ref: [#223](https://git.rtcamp.com/rtmedia/rtMedia/issues/223)


juhi123:
LGTM

PR #1152, opened 2018-06-26
Fix position of magnific popup open in mobile device

malavvasita:
Issue Ref: #1188 

Handle body scroll while maginfic popup is open.


kjnanda:
@malavvasita I cross check this fix at my end. I don't see any difference with scrollable area. 

CC @pranali333

PR #1155, opened 2018-07-05
Add JS hook to handle uploader slide trigger

montu1996:
GL - 229


app/assets/js/rtMedia.backbone.js
```diff
@@ -230,7 +230,9 @@ jQuery( function( $ ) {
 
 						if ( jQuery( 'li#rtm-url-upload' ).length === 0 ) {
 							jQuery( '#' + current_gallery_id + ' .rtmedia-list' ).css( 'opacity', '1' );
-							jQuery( '#rtm-media-gallery-uploader' ).slideUp();
+							if ( rtMediaHook.register( 'rtmedia_js_uploader_slide_after_gallery_reload' ) ) {
```
SID177: `register` requires callback as argument, use `call` to call this hook

Issue #1167, opened 2018-07-13
Issue posting activity update when terms checkbox is enabled.


dharmin:
The issue is only present in buddypress legacy template.

1. Add anything in activity textarea.
2. Now click the "Post Update" button without checking the upload terms checkbox.
3. Now check the upload terms checkbox and click "Post Update".
4. The activity will not be generated, because "Post Update" button is disabled from 2nd step.

Refer the screencast for ref - https://drive.google.com/file/d/14L5R_20H_8gFORhDXxmZH85kcX2eSMgR/view

kjnanda:
The added fix is working as expected.

pooja-muchandikar:
Tested with `develop` branch and works as expected.

pooja-muchandikar:
Tested with `master` branch and works as expected.

PR #1168, opened 2018-07-13
Add remove disable attribute when upload terms checkbox is checked

dharmin:
Enable all form fields disabled which are by BuddyPress on click event of 'Post Update' button
Fix Issue #1167


thrijith:
Fixed in https://github.com/rtMediaWP/rtMedia/commit/30ea5e786f0f2686c3991eb394b2da31b054032a

PR #1172, opened 2018-07-23
Gl 240 uploader shortcode redirect issue

JAshwini:
Redirection code was added in `FileUploaded` because of which the process was getting interrupted, in this PR it is shifted on `ajaxComplete` where there is check if the `UploadComplete` is executed successfully then only redirection code of `ajaxComplete` will be executed.

### DIFF
--- a/app/assets/js/rtMedia.backbone.js
+++ b/app/assets/js/rtMedia.backbone.js
@@ -558,14 +558,17 @@ jQuery( function( $ ) {
 			uploaderObj.initUploader();
 
 			var rtnObj = '';
+			var redirect_request = false;
+
+			jQuery( document ).ajaxComplete(function() {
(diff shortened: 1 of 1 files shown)


app/assets/js/rtMedia.backbone.js
```diff
@@ -557,7 +557,19 @@ jQuery( function( $ ) {
 			uploaderObj = new UploadView( rtMedia_plupload_config );
 			uploaderObj.initUploader();
 
+			var rtnObj = '';
+			var redirect_request = false;
+
+			jQuery( document ).ajaxComplete(function() {
+				if( redirect_request ) {
+					window.location = rtnObj.redirect_url;
```
milindmore22: redirection should be last step

PR #1175, opened 2018-07-26
GL-237: privacy option issue on post activity fixed

JAshwini:
Privacy option was been accessed with reference of the form, so removed that dependency and fetched value through ID as it is same for both buddypress templates.

### DIFF
--- a/app/assets/js/rtMedia.backbone.js
+++ b/app/assets/js/rtMedia.backbone.js
@@ -1413,12 +1413,19 @@ jQuery( document ).ready( function( $ ) {
 				}
 
 				var dynamic_privacy = '';
+				let privacy_element = jQuery( '#rtSelectPrivacy' );
 
-				if ( jQuery( '#whats-new-form select.privacy' ).not( '.rtm-activity-privacy-opt' ).length > 0 ) {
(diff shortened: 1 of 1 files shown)


app/assets/js/rtMedia.backbone.js
```diff
@@ -1413,13 +1413,20 @@ jQuery( document ).ready( function( $ ) {
...
 				var dynamic_privacy = '';
+				let privacy_element = jQuery( '#rtSelectPrivacy' );
 
-				if ( jQuery( '#whats-new-form select.privacy' ).not( '.rtm-activity-privacy-opt' ).length > 0 ) {
-					dynamic_privacy = jQuery( '#whats-new-form select.privacy' ).not( '.rtm-activity-privacy-opt' ).val();
-				} else if ( jQuery( '#whats-new-form input[name="privacy"]' ).length > 0 ) {
-					dynamic_privacy = jQuery( '#whats-new-form input[name="privacy"]' ).val();
+				/**
+				 * Referred ID directly instead of giving reference of form.
+				 * Issue #237
```
SID177: Remove issue comment.

PR #1176, opened 2018-08-01
GL-255: redirect to 404 if media disabled on Profile/Group

JAshwini:


milindmore22:
please fix phpcs issues

PR #1178, opened 2018-08-02
PR - Newly added feature for rtMedia along with new codeception update

kjnanda:
1. Update `composer.json` file to add php version
2. Update `codeception` to latest version. (V 2.4.4)
2. Update selectors in Constants.php file
3. Modify function call in UploadMedia.php, DashboardSettings.php, BuddypressSettings.php
4. Add testcase for `Upload Terms and Privacy Notice`


kjnanda:
@pranali333 JFYI

This will only pass once the privacy settings issue with BP Nouveau template will be fixed.

PR #1203, opened 2018-09-24
Fix incorrect css class applied for media update error.

dharmin:
GL-280 Incorrect CSS class on the error message of media update.


SID177:
Since this PR has conflicts and the changes aren't much, I've added these changes in https://github.com/rtMediaWP/rtMedia/pull/1611.

PR #1206, opened 2018-09-26
Add clear input field code after file input error

dharmin:
GL-283
When the file is selected in settings and if the alert message appears, the file input is not cleared after dismissing the alert box which leads to an error of incorrect file upload.


pooja-muchandikar:
@kapilpaul 

Cross-checked this issue with stable version of rtmedia plugin and found that on uploading incorrect file for upload alert is shown and the name of file is not being saved on closing the alert. 

So this PR is not valid anymore as changes are working as expected in stable version.

Screenshot 2021-10-06 at 5 32 28 PM

Issue #842, opened 2017-03-22
Broken UI for media title (text box) and description (text area)


kjnanda:
**while uploading media as a comment from activity -** 

- [x] 1. For all default theme - text area issue

![broken ui for text area](https://cloud.githubusercontent.com/assets/15173772/24199914/540fedf8-0f31-11e7-995d-7b0677da2b5b.png)

- [ ] 2. Media title ui bug - Specific to Theme 2014 

![broken ui](https://cloud.githubusercontent.com/assets/15173772/24199922/5c31008a-0f31-11e7-9a74-d6650c3f2c04.png)

**While uploading media in group**

- [x] 1. URL -> `/group/Groupname_page`

![broken ui group](https://cloud.githubusercontent.com/assets/15173772/24242079/b35674b6-0fdc-11e7-87b7-17b7a8a2a425.png)

- [x] 2. Please check progress bar UI while uploading the image. 

Check here - https://goo.gl/N3VmkZ

kjnanda:
Please check the fix added in PR - https://github.com/rtMediaWP/rtMedia/pull/1211 and do the needful.

surajkrsingh:
New PR for this issue :- https://github.com/rtMediaWP/rtMedia/pull/1385

kjnanda:
Added changes look good to me. 👍

kjnanda:
@surajkrsingh Changes are not working for me.

kjnanda:
@surajkrsingh Added changes are working as expected with given PR

kjnanda:
This is working as expected and tested with `pre-release` branch. ✅

kjnanda:
This is working as expected and tested with `dev` branch. ✅

kjnanda:
This is working as expected and tested with `master` branch. ✅

kjnanda:
Closing this issue as this has been fixed and release in - V 4.5.8

PR #1211, opened 2018-10-08
Fix media title textbox issue on edit before upload

malavvasita:
Issue Ref: #842 

Checked with all listed issues listed in issue.

Needs testing with different default themes.


kapilpaul:
@pooja-muchandikar as the ref issue was closed stating this was fixed so we can close this PR.

pooja-muchandikar:
Ok closing this PR.

Issue #1006, opened 2017-11-02
Video Page - Broken UI of Pervious button on all default themes


kjnanda:
Please check these screenshots.

**1. Theme 2017**

![2017 - theme](https://user-images.githubusercontent.com/15173772/32310660-fffac8b6-bfb9-11e7-8275-f718387c8dcf.png)
----------------------------------------------------------------------------------------------------
**2. Theme 2016**

![2016 theme](https://user-images.githubusercontent.com/15173772/32310679-1fdd9ad2-bfba-11e7-81ab-2708d05c95f3.png)
----------------------------------------------------------------------------------------------------
3. Theme 2015

![2015 - theme](https://user-images.githubusercontent.com/15173772/32310685-2900bc3e-bfba-11e7-9e57-ca91c4ac5615.png)

thrijith:
@kjnanda  unable to replicate this issue in Wordpress 4.9.1

Theme 2017

![previous2k15](https://user-images.githubusercontent.com/13589980/34296138-140a9b34-e737-11e7-8ee5-ceeb40efb2d0.png)

Theme 2016

![previous2k16](https://user-images.githubusercontent.com/13589980/34296142-1b1911ee-e737-11e7-8c78-c72e28f1ec20.png)

Theme 2015

![previous2k17](https://user-images.githubusercontent.com/13589980/34296152-23def8f2-e737-11e7-8fdb-c1f9fc285036.png)

kjnanda:
Not able to reproduce this issue again. seems to be fixed.

PR #1213, opened 2018-10-09
Fix media view on edit media

malavvasita:
Media was overlapped on Media title.

Fixed by adding margin to media that will provide some space between media and its title.

Issue ref: #1006


pooja-muchandikar:
Closing this PR as this PR changes looks already fixed in stable version.

cc: @kapilpaul

PR #1219, opened 2018-10-15
Add admin ajax upload action

### LINKED ISSUE #1218: Vulnerability on rtUploadAttachment.php
We have discovered several non authorized pdf files on directory  **/httpdocs/wp-content/uploads/rtMedia/tmp** on our website.

They are all about the same size, about 600 Kb and with the same content. Kaspersky and VirusTotal does not detect the pdf as malware. 

On logs we find the same URL invocation from the same IP.

A.B.C.D - - [09/Oct/2018:05:10:11 +0200] "**POST /wp-content/plugins/buddypress-media/app/helper/rtUploadAttachment.php** HTTP/1.0" 200 528 "-" "-"

Every line on log file of this type corresponds to a pdf file with the same timestamp.
WP core and all plugins are updated to the last stable version.

The only way to avoid uploading arbitrary file upload of pdf files has been renaming rtUploadAttachment.php
It is not an elegant solution but the quickest I thought.

Anyone else with this problem?

dharmin:
Fix #1218.
Changed the direct ajax request of rtUploadAttachment.php to admin-ajax.php.


app/helper/rtUploadAttachment.php
```diff
@@ -1,67 +1,92 @@
...
-		$import_export = true;
-	}
+	function rtmedia_admin_upload() {
+		if ( defined( 'DOING_AJAX' ) && DOING_AJAX ) {
+			$nonce = filter_input( INPUT_POST, 'rtmedia_admin_upload_nonce' );
+			if ( wp_verify_nonce( $nonce, 'rtmedia-admin-upload' ) ) {
+				if ( ! empty( $_FILES ) ) {
+					$error = false;
+					$files = array();
+					/* get wp's uploads directory paths and urls. */
```
kiranpotphode: Fix PHPCS issues

PR #1231, opened 2018-11-20
Fix media search with spaces

dharmin:
Fix media search with spaces

GL:309


app/assets/js/rtMedia.backbone.js
```diff
@@ -461,7 +461,8 @@ jQuery( function( $ ) {
 			if ( $( '#search_by' ).length > 0 ) {
 				href += '&search_by=' + $( '#search_by' ).val();
 			}
-
+			href = encodeURI( href );
+			console.log( href );
```
mangesh: `console.log( href );` what is this statement is doing here?

dharmin: Removed the line

PR #1236, opened 2018-11-28
Add missing js data elements

SID177:
Closes https://github.com/rtCamp/rtMedia/issues/311

### DIFF
--- a/templates/main.php
+++ b/templates/main.php
@@ -37,7 +37,7 @@ if ( ! $rt_ajax_request ) {
 			//if it is a buddypress member profile
 			?>
 			<?php do_action( 'bp_before_member_home_content' ); ?>
-		<div id="item-header" role="complementary">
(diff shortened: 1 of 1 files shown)


templates/main.php
```diff
@@ -96,7 +96,7 @@
 	 */
 	do_action( 'bp_before_group_home_content' ); ?>
 
-	<div id="item-header" role="complementary">
+	<div id="item-header" role="complementary" data-bp-item-id="<?php bp_group_id(); ?>" data-bp-item-component="groups" class="groups-header single-headers">
```
abhijitrakas: Could we use `esc_attr` for `bp_group_id()`?

SID177: Its internally escaped, that's why I haven't echoed it here

abhijitrakas: Ok

abhijitrakas:
LGTM! 👍

PR #1239, opened 2018-12-13
Gh 318 redirect issue in shortcode

JAshwini:
Current behavior of shortcode is as follows:

- if we want to redirect to recently uploaded media then we have to specify redirect=true
- if we dont want redirection after media upload then there should be no redirect attribute added to shortcode.

Because of which `redirect=false` was not supporting, so added new paramter `redirection` which will have the value passed in shortcode and if is set to `true` then redirection will be allowed else for any other value it will be false.


JAshwini:
Wrong branch was pushed.

PR #1240, opened 2018-12-13
Gh 318 redirect issue in shortcodes

JAshwini:
Current behavior of shortcode is as follows:

- if we want to redirect to recently uploaded media then we have to specify redirect=true
- if we dont want redirection after media upload then there should be no redirect attribute added to shortcode.

Because of which `redirect=false` was not supporting, so added new paramter `redirection` which will have the value passed in shortcode and if is set to `true` then redirection will be allowed else for any other value it will be false.


app/main/controllers/upload/RTMediaUploadEndpoint.php
```diff
@@ -234,7 +239,7 @@ function template_redirect( $create_activity = true ) {
 				}
 				if ( isset( $this->upload['rtmedia_simple_file_upload'] ) && true == $this->upload['rtmedia_simple_file_upload'] ) {
 					if ( isset( $media ) && count( $media ) > 0 ) {
-						if ( isset( $_redirect_url ) ) {
+						if ( isset( $_redirection ) && 'true' === $_redirection && isset( $_redirect_url ) )  {
```
mangesh: `$_redirection` will contain `string` value or `boolean` walue?

ashwinij1512: String

PR #1241, opened 2018-12-14
Gh 310 album list issue in shortcode

JAshwini:
In gallery shortcode when `uploader=true` is passed along with context value as profile or group, list of albums is shown.

In that if there is no custom album added and only default/global album exists then the markup for album list dropdown was showing nothing in list.

So fixed that if markup is empty then its outer/parent elements should not rendered.


app/main/controllers/upload/RTMediaUploadView.php
```diff
@@ -48,10 +48,18 @@ public function render( $template_name ) {
 			} elseif ( is_rtmedia_album_enable() && $rtmedia_query && is_rtmedia_gallery() ) {
 
 				if ( isset( $rtmedia_query->query['context'] ) && 'profile' === $rtmedia_query->query['context'] ) {
-					$album = '<span> <label> <i class="dashicons dashicons-format-gallery rtmicon"></i>' . esc_html__( 'Album', 'buddypress-media' ) . ': </label><select name="album" class="rtmedia-user-album-list">' . rtmedia_user_album_list() . '</select></span>';
+					$album_list_markup = rtmedia_user_album_list();
+
+					if ( ! empty( $album_list_markup ) ) {
+						$album = '<span> <label> <i class="dashicons dashicons-format-gallery rtmicon"></i>' . esc_html__( 'Album', 'buddypress-media' ) . ': </label><select name="album" class="rtmedia-user-album-list">' . $album_list_markup . '</select></span>';
```
mangesh: Is the output of `$album_list_markup` is escaped?

ashwinij1512: Yes the function `rtmedia_user_album_list` and `rtmedia_group_album_list` functions returns html markup after escaping attributes and text contain.

PR #1247, opened 2018-12-25
Gh 317 public message media not displayed

JAshwini:
Initially activity type and content was getting updated directly in database on `bp_activity_posted_update` action, but because of this the Ajax request was not getting updated, so added filter `bp_activity_type_before_save` and `bp_activity_content_before_save`.

### DIFF
--- a/app/main/controllers/activity/RTMediaBuddyPressActivity.php
+++ b/app/main/controllers/activity/RTMediaBuddyPressActivity.php
@@ -204,12 +204,13 @@ class RTMediaBuddyPressActivity {
 	 * This function will check for the media file attached to the actitvity and accordingly will set type.
 	 *
 	 * @param string $type Type of the Activity.
-	 * @return string      Filtered value of the activity type.
+	 *
+	 * @return string Filtered value of the activity type.
 	 */
 	public function bp_activity_type_before_save( $type ) {
 
 		$rtmedia_attached_files = filter_input( INPUT_POST, 'rtMedia_attached_files', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY );
-		if ( is_array( $rtmedia_attached_files ) ) {
+		if ( ( ! empty( $rtmedia_attached_files ) ) && is_array( $rtmedia_attached_files ) ) {
 			$type = 'rtmedia_update';
 		}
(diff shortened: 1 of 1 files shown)


app/main/controllers/activity/RTMediaBuddyPressActivity.php
```diff
@@ -198,6 +200,37 @@ function bp_get_activity_parent_content( $content ) {
...
+	/**
+	 * This function will check for the media file attached to the actitvity and accordingly will set type.
+	 *
+	 * @param string $type Type of the Activity.
+	 * @return string      Filtered value of the activity type.
+	 */
+	public function bp_activity_type_before_save( $type ) {
+
+		$rtmedia_attached_files = filter_input( INPUT_POST, 'rtMedia_attached_files', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY );
+		if ( is_array( $rtmedia_attached_files ) ) {
```
abhijitrakas: Could we use `! empty` here? check all other places.

app/main/controllers/activity/RTMediaBuddyPressActivity.php
```diff
@@ -198,6 +200,37 @@ function bp_get_activity_parent_content( $content ) {
 		return $content;
 	}
 
+	/**
+	 * This function will check for the media file attached to the actitvity and accordingly will set type.
+	 *
+	 * @param string $type Type of the Activity.
```
abhijitrakas: Maintain alignment after param type.

app/main/controllers/activity/RTMediaBuddyPressActivity.php
JAshwini: Sure :+1:

app/main/controllers/activity/RTMediaBuddyPressActivity.php
JAshwini: Done :+1:

app/main/controllers/activity/RTMediaBuddyPressActivity.php
JAshwini: Done :+1:

Issue #1248, opened 2018-12-31
[Customer Query - [#2493000]] rtmedia_no_media_found_message_filter is not working


kjnanda:
Please look into this query.

Check this [link](https://rtmedia.io/ticket/2493000/#followup-281712) for more information.

kjnanda:
@JAshwini This is working as expected now.

kjnanda:
@ashwinij1512 Please do not merge this in `dev-branch`. I need to test some edge cases.

kjnanda:
@ashwinij1512 Tested at my end with looks good to me. 👍

pooja-muchandikar:
Release in version `4.5.5`

PR #1252, opened 2019-01-03
GH-1248 - Fixed rtmedia_no_media_found_message_filter content echo issue

JAshwini:


templates/media/album-gallery.php
```diff
@@ -72,7 +72,8 @@
 	<?php } else { ?>
 		<p class="rtmedia-no-media-found">
 			<?php
-			apply_filters( 'rtmedia_no_media_found_message_filter', esc_html_e( 'Sorry !! There\'s no media found for the request !!', 'buddypress-media' ) );
+			$message = apply_filters( 'rtmedia_no_media_found_message_filter', 'Sorry !! There\'s no media found for the request !!' );
+			echo esc_html__( $message, 'buddypress-media' );
```
SID177: @JAshwini You can use `esc_html_e` instead. Please check at all instances.

JAshwini: @SID177 
For following syntax `esc_html_e()` and `echo esc_html__()`, there is no difference, either of them can be used so there is no need to replace.

PR #1298, opened 2019-03-06
Fix issue for enable/disable of likes tab

surajkrsingh:
Issue : Create likes tab automatically based on allowedType_likes_enable attribute but it should be create through addons.

Solution: Just skip the allowedType_likes_enable attribute to create like tab.


surajkrsingh:
This issue is already solved in rtmedia-likes.
Note:- Now closing the PR because solution is already merge in develop branch within rtmedia-likes so this solution may conflict .

Issue #1301, opened 2019-03-12
Title attribute is missing for media likes


pooja-muchandikar:
Like we have title attribute for comment likes, we can have title attribute for media likes as well.

Refer screenshot where we have title attribute for comments like, but for media likes title attribute is not there.
https://drive.google.com/open?id=1VjYqDXte0J7sx9dTzULdYJ6xgGhHo8IU
##

Refer below screenshot of comment like where title attribute is added, similarly we need to add it for media likes.
https://drive.google.com/open?id=10-EJXzAQZfx-LQmDo5zGa0WWcSu-nTGd
##

surajkrsingh:
PR:- https://github.com/rtMediaWP/rtMedia/pull/1302

pooja-muchandikar:
@surajkrsingh 

Tested and changes are done as expected. Title attribute for Media Likes is shown now. :heavy_check_mark:

kjnanda:
Tested with dev branch. Added changes are working as expected.

kjnanda:
Tested with master branch. Added changes are working as expected.

PR #1302, opened 2019-03-12
Add title attribute media like option

surajkrsingh:
Issue ref:- https://github.com/rtMediaWP/rtMedia/issues/1301

Issue :- tool-tip is not visible when we hover on like option.
Solution :- Add title attribute in like option on tag as well as in rtMedia.backbone.js on ajax function.

### DIFF
--- a/app/main/controllers/media/RTMediaUserInteraction.php
+++ b/app/main/controllers/media/RTMediaUserInteraction.php
@@ -215,9 +215,9 @@ class RTMediaUserInteraction {
 				$icon = "<i class='" . esc_attr( $this->icon_class ) . "'></i>";
 			}
 			$button_start = '<form action="' . esc_url( $link ) . '">';
-			$button_label = apply_filters( 'rtmedia_' . $this->action . '_label_text', esc_html( $this->label ) );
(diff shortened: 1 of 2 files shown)


app/main/controllers/media/RTMediaUserInteraction.php
```diff
@@ -215,8 +215,9 @@ function render() {
 				$icon = "<i class='" . esc_attr( $this->icon_class ) . "'></i>";
 			}
 			$button_start = '<form action="' . esc_url( $link ) . '">';
+			$button_label = apply_filters( 'rtmedia_' . $this->action . '_label_text', esc_html( $this->label ) );
```
thrijith: @surajkrsingh final value of `$button_label` is being escaped before use. Is `esc_html` necessary here ( double escaping ) ?

surajkrsingh: @thrijith  I think it is not required to escape early.

surajkrsingh: Resolved it.

Issue #1288, opened 2019-03-01
Should not display <span>&nbsp;</span> when no comment has been made while uploading media as comment


kjnanda:
> **Migration Note:** This issue was originally created by @kjnanda 

Steps:

1. Open any media in Lightbox or (in Single Media page)
2. Upload media as in comment without adding any text. (Please Enable `Enable media in comment` setting in Buddypress tab)
3. Check on Activity Page for the newly added activity of comment made on media.(Please Enable this setting `Create an activity for media comments` in Buddpyress tab)
4. It will show up space for text line. It should not display when no text comment has been added. 

`Note: You can check the same thing from activity page also`

-----

Please refer this snapshot.
![krupa___test](https://user-images.githubusercontent.com/15173772/53624072-f5b0ba80-3c24-11e9-9097-c48e5c3ad902.png)

surajkrsingh:
PR:- https://github.com/rtMediaWP/rtMedia/pull/1305

pooja-muchandikar:
@surajkrsingh 

Testing is done and changes done are not reflecting.
1. Cross-checked uploading media from activity page without any text and in activity feed <span>$nbso;</span> is still shown.
2. Checked media comment activity as well and the issue still exists. 

Refer screenshot of activity generated after media commenting: 
![comment](https://user-images.githubusercontent.com/41000648/54512461-4856da00-497a-11e9-90db-e3dc31892e92.jpeg)
##

Refer screenshot of activity posted from activity page itself:
![activity-page](https://user-images.githubusercontent.com/41000648/54512487-5d336d80-497a-11e9-87cc-b88b14e00413.jpeg)
##

```
Test Environment
WordPress: v5.1.1
rtMedia: Code from the pre-release branch v4.5.6
BuddyPress: v4.2.0 ( Nouveau and Legacy template )
Themes: 2012, 2016, 2017 and 2019
```

surajkrsingh:
PR:-https://github.com/rtMediaWP/rtMedia/pull/1307

pooja-muchandikar:
@surajkrsingh 

Tested and changes done looks fine for **media uploaded without text on Activity page :heavy_check_mark:** for comments activity generated unable to cross-check as commenting media file from lightbox is generating media activity twice because of which unable to verify &nbsp issue for comments activity. Will revisit this issue once comment issue is fixed.  
Ref link: https://github.com/rtMediaWP/rtMedia/issues/1285
```
Test Environment
[...]

PR #1305, opened 2019-03-15
Remove extra span from media comment

surajkrsingh:
Issue ref:- https://github.com/rtMediaWP/rtMedia/issues/1288

Issue: When we upload a media without comment a text then empty span is added to the comment .

Solution:- Add logical condition to check for comment text is empty or not and then stored data via ajax call.

### DIFF
--- a/app/assets/js/rtMedia.backbone.js
+++ b/app/assets/js/rtMedia.backbone.js
@@ -1689,7 +1689,11 @@ jQuery( document ).ready( function( $ ) {
 		$( this ).attr( 'disabled', 'disabled' );
 
 		// Sanitize comment content and escape html tags
-		comment_content_el.val( jQuery( '<span/>' ).text( jQuery.trim( comment_content_el.val() ) ).html() );
+		if( '' !== comment_content_el.val().replace(/\&nbsp;/g, '') ) {
+			comment_content_el.val( jQuery( '<span/>' ).text( jQuery.trim( comment_content_el.val() ) ).html() );
+		} else {
+			comment_content_el.val( '' );
+		}
 
 		$.ajax( {
 			url: comment_form_el.attr( 'action' ),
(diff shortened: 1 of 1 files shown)


app/assets/js/rtMedia.backbone.js
```diff
@@ -1689,7 +1689,11 @@ jQuery( document ).ready( function( $ ) {
 		$( this ).attr( 'disabled', 'disabled' );
 
 		// Sanitize comment content and escape html tags
-		comment_content_el.val( jQuery( '<span/>' ).text( jQuery.trim( comment_content_el.val() ) ).html() );
+		if( '' !== comment_content_el.val().replace(/\&nbsp;/g, '') ) {
+			comment_content_el.val( jQuery( '<span/>' ).text( jQuery.trim( comment_content_el.val() ) ).html() );
```
chandrapatel: Can't you use `comment_content_el.append()`?

`comment_content_el.append( jQuery( '<span/>' ).text( jQuery.trim( comment_content_el.val() ) ) );`

app/assets/js/rtMedia.backbone.js
```diff
@@ -1689,7 +1689,11 @@ jQuery( document ).ready( function( $ ) {
 		$( this ).attr( 'disabled', 'disabled' );
 
 		// Sanitize comment content and escape html tags
-		comment_content_el.val( jQuery( '<span/>' ).text( jQuery.trim( comment_content_el.val() ) ).html() );
+		if( '' !== comment_content_el.val().replace(/\&nbsp;/g, '') ) {
```
chandrapatel: Left space before open `(`

`if ( '' !== comment_content_el.val().replace(/\&nbsp;/g, '') ) {`

app/assets/js/rtMedia.backbone.js
chandrapatel: Avoid `.html()` to get/set content.

surajkrsingh: We can use append()  also but I think both the way is correct.
Example:-
input - 'hello'
output : &lt;span&gt;hello &lt;/span&gt;
According to my view .val() is better one because we just replacing value not appending a new within it.

Issue #1292, opened 2019-03-01
[Upload terms checkbox] Issue list when direct upload and upload terms kept enabled together


kjnanda:
Please check my comment below.

- [ ] 1.

## ~~Scenario 1 - First add some text and click on upload terms checkbox~~ 

It is not giving me time to upload the media as soon as I click on `upload terms` it posts the text area content.

Please refer this link for visual - https://goo.gl/13XtLf

-----

- [ ] 2.

## Scenario 2 - First upload some media and then try to enter some text in text area

Text area gets disabled and it is not allowing the user to enter anything. 

Please refer this visual - https://goo.gl/duvckW

-----

```
Test Environment : 

1. WP 5.1
2. Buddypress 4.2.0
3. rtMedia core - 4.5.6
```

surajkrsingh:
@kjnanda 
For the scenario-1
For posting something from activity both the things ( comment and media ) are not necessary so it quickly post when we click on t&c checkbox.

For the scenario-2
Textarea is not disable just added &nbsp; and the color of content is white so it not visible to us and look like disable.
you can write on textarea but it not visible because it's contents are white same as bg of textarea so after post you can see the contents.

pooja-muchandikar:
@surajkrsingh 
Cross-checked both the scenarios at my end and below are my findings. 
**For Scenario-1:** - yes its most likely that user might want to enter text and media in the activity feed. If wants to then, the user might add both contents ( media and text ) and then click on upload terms checkbox.
##

**For Scenario -2** - textarea doesn't get disabled but the text entered in the textarea is not visible and hence need to fix this issue. 
##

cc: @kjnanda

surajkrsingh:
PR: - https://github.com/rtMediaWP/rtMedia/pull/1306

surajkrsingh:
@pooja-muchandikar  @kjnanda 
In above issue the buddypress doesn't allow to keep empty to the textarea when we post something from activity post.
So before generated this issue we added space with transparent color so it looks like nothing in textarea (and disable too) so this thing create issue when user want to add something in textarea after upload a media.
For fixing this issue remove the white space before submit so user can enter some text and it doesn't look like disable

Note:- If user doesn't enter any text in textarea then after click on sumbit we forcefully put the white space so we can post.

pooja-muchandikar:
@surajkrsingh 
>For Scenario-1:
[...]

PR #1306, opened 2019-03-18
Fix issue acivity-upload-post

surajkrsingh:
Issue ref:- https://github.com/rtMediaWP/rtMedia/issues/1292

Issue:- After upload a media in activity post user can not comment text and  textarea is like disable added string &nbsp;

After solution:- Even after upload media user can comment some text.


app/assets/js/rtMedia.backbone.js
```diff
@@ -1292,11 +1292,8 @@ jQuery( document ).ready( function( $ ) {
 				 * if user want media and activity_text both require
 				 * By: Yahil
 				 */
-				if ( jQuery.trim( jQuery( "#whats-new" ).val() ) == "" ) {
-					if ( rtmedia_activity_text_with_attachment == 'disable') {
-						$( "#whats-new" ).css( 'color', 'transparent' );
-						$( "#whats-new" ).val( '&nbsp;' );
-					} else {
+				if ( '' === jQuery.trim( jQuery( "#whats-new" ).val() )) {
```
SID177: Please add single space before closing bracket `)` at the end.

surajkrsingh: resolved it.

PR #1313, opened 2019-03-19
 Fix issue for direct upload media with content in legacy template pack 

surajkrsingh:
Issue ref:- https://github.com/rtMediaWP/rtMedia/issues/1292

Issue:- After resolve the issue of media upload without content in activity post conflict when term & condition checkbox is disable within legacy template pack.


surajkrsingh:
Get more conflict with many changes so now closing this.

Issue #1284, opened 2019-02-25
[Theme specific] Issue with 2017 - Media is not opening in light box


kjnanda:
After updating theme 2017 to its latest version ( 2.0 --> 2.1), media is not opening in lightbox even if the lightbox setting is kept on from rtMedia settings.

`Expected Result:` Clicking upon media - it should open in lightbox

`Actual Result:` Clicking upon media it takes the user to media single page

**Note: The issue appears only whie opening media from `Media Page`. The same media is opening in lightbox from actvity page.**

-----

```
Test Environment:

1. WP - 5.1 and 5.0.3
2. Buddypress  - 4.2.0 and 4.1.0
3. rtMedia - 4.5.5 and 4.5.6
4. Themes - 2013 to 2019 - But issue is present in latest 2017 theme only
5. Browser - Chrome, firefox and safari
```

surajkrsingh:
PR:-https://github.com/rtMediaWP/rtMedia/pull/1315

pooja-muchandikar:
@surajkrsingh 

Testing is done, and the changes added works fine. Able to view media in the lightbox in 2017 theme. 
```
Test Environment
WordPress: v5.1.1
rtMedia: Code from **pre-release** branch (v4.5.6)
BuddyPress: v4.2.0 ( Nouveau and Legacy template )
Themes: 2017 theme as the issue was theme specific. Also cross-checked 2012, 2016 and 2019 theme
```

kjnanda:
Media can be opened in lightbox now. TEsted this against dev branch.

kjnanda:
Media can be opened in light box now. Tested this against master branch.

PR #1315, opened 2019-03-19
 Fix media lightbox issue for theme-2017 v2.1

surajkrsingh:
Issue ref:- https://github.com/rtMediaWP/rtMedia/issues/1284

Issue :- Media was not opening in lightbox in theme-2017 v2.1 because conflict class has-sidebar.

Solution:-Remove that particular class in single media page in which sidebar is not exist.


app/assets/js/rtMedia.js
```diff
@@ -1290,3 +1290,12 @@ jQuery( document ).ready( function () {
 	// Call function when document loaded.
 	rtm_upload_terms_activity();
 });
+
+jQuery( 'document' ).ready( function() {
+	/** Avoid Lightbox conflict due to class has-slider added by theme 2017 v2.1 */
```
thrijith: Use `// ` for single line comment.

app/assets/js/rtMedia.js
```diff
@@ -1290,3 +1290,12 @@ jQuery( document ).ready( function () {
...
+	if( jQuery( 'body' ).hasClass( 'has-sidebar' ) && 0 === jQuery( '#secondary' ).length ) {
```
thrijith: There is a `.ready(` available above, can we move this check there?

PR #1317, opened 2019-03-21
Fix/deleted media thumbnail

kanumalivad:
every media url has unique timestamp so it won't be conflict with cache image.

### DIFF
--- a/app/main/controllers/template/rtmedia-functions.php
+++ b/app/main/controllers/template/rtmedia-functions.php
@@ -692,9 +692,9 @@ function rtmedia_image( $size = 'rt_media_thumbnail', $id = false, $recho = true
 	$src = apply_filters( 'rtmedia_media_thumb', $src, $media_object->id, $media_object->media_type );
 
 	if ( true === $recho ) {
-		echo set_url_scheme( $src ). '?' .time();
+		echo set_url_scheme( $src ) . '?' . time();
 	} else {
-		return $src. '?' .time();
+		return $src . '?' . time();
 	}
 
 }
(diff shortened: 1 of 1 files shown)


app/main/controllers/template/rtmedia-functions.php
```diff
@@ -692,9 +692,9 @@ function rtmedia_image( $size = 'rt_media_thumbnail', $id = false, $recho = true
 	$src = apply_filters( 'rtmedia_media_thumb', $src, $media_object->id, $media_object->media_type );
 
 	if ( true === $recho ) {
-		echo set_url_scheme( $src );
+		echo set_url_scheme( $src ) . '?' . time();
```
thrijith: @kanumalivad Please add comment why the timestamp was added, for future references.

kanumalivad: okk

app/main/controllers/template/rtmedia-functions.php
```diff
@@ -691,10 +691,13 @@ function rtmedia_image( $size = 'rt_media_thumbnail', $id = false, $recho = true
...
 	$src = apply_filters( 'rtmedia_media_thumb', $src, $media_object->id, $media_object->media_type );
 
+	//Added timestamp because conflict with cache image.
+	$src = $src . '?' . time();
+
 	if ( true === $recho ) {
 		echo set_url_scheme( $src );
 	} else {
-		return $src;
+		return $src.
```
thrijith: @kanumalivad it should be `;` instead of `.`
Check build failure [here](https://travis-ci.org/rtMediaWP/rtMedia/builds/509382968#L1158)

PR #1318, opened 2019-03-22
Fix/wrong media is displayed in lightbox

kanumalivad:
add timestamp at the end of url.


app/main/controllers/template/rtmedia-filters.php
```diff
@@ -282,6 +282,9 @@ function replace_src_with_transcoded_file_url( $html, $rtmedia_media ) {
 		$final_file_url = wp_get_attachment_url( $attachment_id );
 	}
 
+    //Add timestamp to resolve conflict with cache media.
+    $final_file_url = $final_file_url . '?' . time();
+    
 	return preg_replace( '/src=["]([^"]+)["]/', "src=\"$final_file_url\"", $html );
```
thrijith: Since `'?' . time()` has to be added to the src anyway, can we add it directly to `src` attribute?

app/main/controllers/template/rtmedia-functions.php
```diff
@@ -543,6 +543,9 @@ function rtmedia_media( $size_flag = true, $echo = true, $media_size = 'rt_media
 		if ( 'photo' === $rtmedia_media->media_type ) {
 			$src  = wp_get_attachment_image_src( $rtmedia_media->media_id, $media_size );
 
+			//Addtime stamp to resolve conflict with cache image.
+			$src[0] = $src[0] . '?' . time();
```
thrijith: Same as above add `'?' . time();` in `set_url_scheme` usage of `$html`

thrijith: >Add timestamp to resolve conflict with cache image.

add this as the last line of multiline comment, to indicate addition of timestamp.

app/main/controllers/template/rtmedia-filters.php
kanumalivad: yes, added.

PR #1323, opened 2019-04-04
Improve UI of options list

kanumalivad:
issue ref :- https://github.com/rtCamp/rtmedia-favorites/issues/11


app/assets/css/sass/_popup.scss
```diff
@@ -151,6 +143,14 @@
 			color: $white;
 			padding: 0;
 		}
+
+		form button {
```
ravatparmar: Suggestion: If the button has any class can we use it here, otherwise in future anyone will add a button inside this form it will have this stylesheet applied.

kanumalivad: @ravatparmar this UI improvements needs to apply for all buttons not only the button which is added from the add on. 

that's why i changed it in core css.

Issue #1291, opened 2019-03-01
Issue with load more option when number of media on page is set to 1


kjnanda:
Please check the steps to reproduce issues.

1. Goto Display tab under rtMedia settings.
2. Set some value for `Number of media per page`. Eg. We have kept it `1`.
3. Select `Load more` option.
4. Click on `Save Settings` button.
5. Goto media page and open any media in light box.
6. Click on `X` button and check one extra media gets appended as shows in visual.

By default `Load more` button gets clicked. 🤔

--------

Please check this visual - https://goo.gl/pSsARF

-----

```
Test Environment Used.

1. rtMedia V4.5.6
2. BP V4.2.0
3. WP V5.1
4. Browser Chrom, Firefox and Safari.
```

pooja-muchandikar:
@kanumalivad 

Testing is done with the following test environment.
```
Test Environment
rtMedia Core: v4.5.6 ( code from **pre-release** branch )
BuddyPress: v4.2.0 ( nouveau and legacy template )
Themes: Default themes 2012, 2016 and 2019 
```

The fix added for the issue is working but also found a regression bug ( UI glitch ). Follow the steps to reproduce the regression bug:
1. Go to rtMedia settings > Display tab > and set media per page to 5 and select load more option
2. Go to media page in the frontend
3. Click on the last media on the page, media opens in lightbox.
4. Now, close the lightbox, observe that on closing lightbox UI glitch is visible. Blank space is shown between the media and the load more button.

Refer: https://drive.google.com/file/d/1kUP-N91xIyltv0rCNKKxPf-Vbm6jJXmj/view

Also, when viewing media in lightbox next 5 media will be loaded in the background of the lightbox this issue occurs sometimes.

pooja-muchandikar:
@kjnanda 

Can you cross check comment https://github.com/rtMediaWP/rtMedia/issues/1291#issuecomment-484452633 and confirm if you are facing this issue?

PR #1327, opened 2019-04-12
Fix/load more issue

kanumalivad:
issue ref : - https://github.com/rtMediaWP/rtMedia/issues/1291
In lighbox when last or last second media is opened it will load more media to slide and then if we close the lightbox, In media gallery it will show loaded media also that shouldn't.

So here when more media will load i take some values and after lightbox close i set them as before.


app/assets/js/rtMedia.js
```diff
@@ -143,6 +154,21 @@ function apply_rtMagnificPopup( selector ) {
...
+						//If more media is loaded in lighbox then remove them set nextpage to default one.
+						if ( more_media_loaded ) {
+
+							mfp.ev.empty();
+							mfp.ev.append( old_gallery_media );
+
+							nextpage          = current_page;
+							more_media_loaded = false;
+
+							if( nextpage > 1 ) {
```
ravatparmar: Put one space after if.

app/assets/js/rtMedia.js
```diff
@@ -11,6 +11,10 @@ function apply_rtMagnificPopup( selector ) {
 		}
 		if ( typeof( rtmedia_lightbox_enabled ) != 'undefined' && rtmedia_lightbox_enabled == '1' ) { // If lightbox is enabled.
 
+			var old_gallery_media;
+			var current_page;
```
ravatparmar: Please check if we need to initialize variable because only declared variable will have "undefined" value.

kanumalivad: I have checked, There is no need to intialize when declaring.

Issue #1319, opened 2019-03-22
Group Activity media are not shown in Group Media tab with BuddyPress Nouveau Template.


pooja-muchandikar:
**Steps to Reproduce**

1. Set BuddyPress template to Nouveau and Save settings
2. Go to Groups and create a group
3. Go to group activity and upload any media in the group activity
4. Observe that media uploaded through group activity are not reflecting in the group media tab.
5. Now, go to profile media tab and observe that media uploaded through group activity are reflecting in the profile media tab which shouldn't be.

**_Media uploaded through group activity should reflect in the group media tab, not in the profile media tab. This issue exists with the nouveau template. With BuddyPress legacy template it is working as expected._**

**Refer Visual:** https://drive.google.com/open?id=1OtUZFW6nuhIbxJkGKGXkZUKC4ZiZA2og
##

```
Test Environment
WordPress: 5.1.1
rtMedia: 4.5.6
BuddyPress: 4.2.0 ( Nouveau Template )
```
cc: @kjnanda

pooja-muchandikar:
@kanumalivad 

Testing is done with the following test environment and fix added works as expected. 

```
Test Environment
rtMedia Core: v4.5.6 ( code from **pre-release** branch )
BuddyPress: v4.2.0 ( nouveau and legacy template )
```

kjnanda:
Tested with dev branch. Added changes are working as expected.

kjnanda:
Tested with master branch. Added changes are working as expected.

PR #1331, opened 2019-04-16
Add condition for nouveau template

kanumalivad:
issue ref:- https://github.com/rtMediaWP/rtMedia/issues/1319
Here item id is get by `#whats-new-post-in` id which is not avilable in nouveau template.
I found there is `.groups-header` class applied on tag which has group id.

### DIFF
--- a/app/assets/js/rtMedia.backbone.js
+++ b/app/assets/js/rtMedia.backbone.js
@@ -1394,11 +1394,15 @@ jQuery( document ).ready( function( $ ) {
  				}
 			} );
 
-			var object = '';
-			var item_id = jQuery( '#whats-new-post-in' ).val();
-			if ( item_id == undefined ) {
-				item_id = 0;
- 			}
+			var object  = '';
+			var item_id = 0;
+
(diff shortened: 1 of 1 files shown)


app/assets/js/rtMedia.backbone.js
```diff
@@ -1394,11 +1394,15 @@ jQuery( document ).ready( function( $ ) {
...
-			var item_id = jQuery( '#whats-new-post-in' ).val();
-			if ( item_id == undefined ) {
-				item_id = 0;
- 			}
+			var object  = '';
+			var item_id = 0;
+
+			if ( jQuery( '#whats-new-post-in' ).length ) {
+				item_id = jQuery( '#whats-new-post-in' ).val();
+			} else if ( jQuery( '.groups-header' ).length ) {
```
ravatparmar: Please check if we also need to put a check for the current page is group or not.

kanumalivad: checked it, no need to put check for curre page is group or not.

PR #1335, opened 2019-04-23
fix alignment issue in water mark color setting

mibhumi:
issue reference: https://github.com/rtCamp/rtMedia/issues/354


app/assets/admin/css/sass/_components.scss
```diff
@@ -72,8 +72,7 @@
 			vertical-align: top;
 
 			fieldset label {
-				display: block;
-				margin: 0 !important;
+				margin: 0 6px 0 0 !important;
```
surajkrsingh: Please try to avoid the use of `! important`.

mibhumi: I haven't add `!important`. I had just removed `display: block.` and give margin of right side.

surajkrsingh: But you have made changes in this line so as I see ` ! important ` is still there please try to remove if possible.

Issue #367, opened 2013-07-09
No Privacy in media edit section


nitun:
We need to give option to change uploaded media to make it private or users or public,

PR #1345, opened 2019-05-16
[GH-367] Fix multiple media upload from activity for non-admin users

dharmin:
Added filter to bypass the limit of links on activity by changing the option of `comment_max_links`.

### DIFF
--- a/app/main/controllers/template/rtmedia-filters.php
+++ b/app/main/controllers/template/rtmedia-filters.php
@@ -816,20 +816,6 @@ function rtmedia_model_query_columns( $columns ) {
 
 add_filter( 'rtmedia-model-query-columns', 'rtmedia_model_query_columns', 10, 1 );
 
-function rtmedia_comment_max_links_callback( $values, $option = '' ) {
-	$new_values = $values;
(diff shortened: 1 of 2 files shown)


app/main/controllers/template/rtmedia-filters.php
```diff
@@ -816,20 +816,6 @@ function rtmedia_model_query_columns( $columns ) {
...
-		$rtmedia_attached_files = filter_input( INPUT_POST, 'rtMedia_attached_files', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY );
-		if ( is_array( $rtmedia_attached_files ) && ! empty( $rtmedia_attached_files[0] ) ) {
-			if ( $new_values < 5 ) {
-				$new_values = 5;
-			}
-		}
-	}
-	return $new_values;
-}
-add_filter( 'option_comment_max_links', 'rtmedia_comment_max_links_callback', 10, 2 );
```
thrijith: @dharmin Is this filter not used? Why is it removed?

dharmin: Updated 'pre_option_*' to 'option_*'.

Issue #1285, opened 2019-02-25
Media uploaded in comments from media lightbox is shown twice in activity feed.


pooja-muchandikar:
```
Test Environment
rtMedia Core: v4.5.5 and v4.5.6
BuddyPress: v4.1.0 and v4.2.0
Theme: Default theme
```
**Steps to Reproduce**
1. Go to rtMedia Settings > BuddyPress and enable `Create an activity for media comments`  ( check disabling this option as well )
2. Go to activity page, and post an update
3. Now, comment on the post with and without media, and observe that without media comment activity is generated correctly, and also with media comment activity is generated correctly.
4. Now, Go to Media page and open media in lightbox ( check media from activity page also )
5. Add text comment to the media from the lightbox, and observe in activity page comment activity is generated correctly.
6. Now, add media comment to the media from the lightbox, and observe in activity page comment activity is generated and media uploaded as a comment is shown twice, which is incorrect. 

Refer this visual: https://drive.google.com/open?id=1_tEItOt_GoUz7PHC7anW1rBmO_uPYucc

**Note: Issue exists in Groups as well.**

surajkrsingh:
@kjnanda @dharmin @sanketio 
I have added the solution to fix the duplication of commented media on activity.

kjnanda:
Tested this issue with pre-release branch and this is working as expected now.

kjnanda:
@dharmin With the latest code from dev and pre-release branch, it introduced a `Regression issue.`

**Please read the scenario:**

1. Uploading the media as comment from light box - Works as expected ✅ 

2. Uploading the media as comment from activity - It shows duplicate entry for the respective uploaded media on activity page itself. ❌ 

Note that when I disable the `Create activity for media comments` issue no more appears. 

------

Please check the cause and try to fix it asap.

We have already discussed the issue. though let me know if you need the visual for this.

kjnanda:
Tested with dev branch. Added changes are working as expected.

https://github.com/rtMediaWP/rtMedia/issues/1285#issuecomment-500435848 Both testcases are working as expected now.

kjnanda:
Test with master branch and it is working as expected.

PR #1352, opened 2019-06-03
 Fix duplication of comment media on activity 

surajkrsingh:
Issue ref: #1285

### DIFF
--- a/app/main/controllers/activity/RTMediaBuddyPressActivity.php
+++ b/app/main/controllers/activity/RTMediaBuddyPressActivity.php
@@ -259,6 +259,12 @@ class RTMediaBuddyPressActivity {
 	 */
 	public function bp_activity_content_before_save( $content ) {
 
+		// Fix duplication issue of comment media on activity by returning contents when activity type is comment.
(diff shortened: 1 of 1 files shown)


app/main/controllers/activity/RTMediaBuddyPressActivity.php
```diff
@@ -259,6 +259,12 @@ function bp_get_activity_parent_content( $content ) {
 	 */
 	public function bp_activity_content_before_save( $content ) {
 
+		// Fix duplication issue of comment media on activity by returning contents when activity type is comment.
+		$rtmedia_comment_nonce = filter_input( INPUT_POST, 'rtmedia_comment_nonce', FILTER_SANITIZE_STRING );
+		if ( ! empty( $rtmedia_comment_nonce ) ) {
```
sanketio: Why are we just checking the nonce if it is empty or not instead of verifying? Any reason? Am I missing anything?

cc: @dharmin @surajkrsingh

dharmin: @sanketio 
The code below the condition should need to run if it's not commented. So need to check that the activity which is created is from comment, and the nonce was only feasible field to check if this block is run after posting the comment.

@surajkrsingh Please correct me if something I wrote is incorrect.

PR #1362, opened 2019-06-11
[GH-367] Fix multiple media upload from activity for non-admin users

dharmin:
Added filter to bypass the limit of links on activity by changing the option of `comment_max_links.`


kjnanda:
@dharmin Tested this PR and it works fine

PR #1390, opened 2019-06-21
Fix media title thumbnail overlapping issue on legacy template

dharmin:
Override some of the default styles of MediaElementJs and BuddyPress.


app/assets/css/sass/_activity.scss
```diff
@@ -73,10 +73,20 @@
...
+    ul.rtmedia-list {
+        li {
+            border-bottom: 0;
+            padding-bottom: 0;
+        }
+    }
+    // Media title fix on legacy .
+    ul.activity-list {
+        li {
+            overflow: unset!important;
```
thrijith: @dharmin Can we avoid usage of `! important` here? If possible override by adding more specificity.

dharmin: BuddyPress has inline CSS control over it, so need to add `!important`.

Issue #1376, opened 2019-06-12
Tooltip Alignment issue with watermark add-on settings


kjnanda:
```
Test Environment 
rtMedia Core: v4.5.6
BuddyPress: v4.1.0
rtMedia Photo watermark add-on: code from pre-release branch ( current version 1.2.5 ) 
```
1. Go to rtMedia Settings tab
2. Navigate to watermark settings page.
3. Go to font color and click on select color.
4. Observe that, when color picker is displayed, the `info` icon gets misaligned. Also, the `clear` button is also misaligned.

Refer below snapshot.
![ui-issue](https://user-images.githubusercontent.com/41000648/53093860-98d74500-353e-11e9-991d-b2d581229e08.png)

kjnanda:
@surajkrsingh Please look into this issue

surajkrsingh:
PR:-https://github.com/rtMediaWP/rtMedia/pull/1405/files

kjnanda:
This is working as expected and tested with `pre-release` branch. ✅

kjnanda:
This is working as expected and tested with `pre-release` branch. ✅

kjnanda:
This is working as expected and tested with `dev` branch. ✅

kjnanda:
This is working as expected and tested with `master` branch. ✅

kjnanda:
Closing this issue as this has been fixed and release in - V 4.5.8

PR #1405, opened 2019-07-17
Fix position for watermark color picker clear button

surajkrsingh:
Issue ref:- https://github.com/rtMediaWP/rtMedia/issues/1376


app/assets/admin/css/sass/_components.scss
```diff
@@ -82,8 +82,18 @@
 				margin-top: 5px !important;
 			}
 		}
+
+		.wp-picker-input-wrap label {
```
thrijith: Wrap `label` css in `.wp-picker-input-wrap { }`

Check: https://github.com/rtMediaWP/rtMedia/pull/1405/files#diff-1a760e790d6d44ec167aaa0b34bc8c60L81

surajkrsingh: Done

Issue #1432, opened 2019-10-08
Issue list with BuddyPress 5.0.0


pooja-muchandikar:
- [x] 1. Post update button on Activity page is not visible with BP 5.0.0 with Nouveau template. 
```
Test Environment
WordPress: v5.2.3
rtMedia: v4.5.8
BuddyPress: 5.0.0 ( Nouveau Template )
Theme: 2019, 2017
```

dishitpala:
_1. Post update button on Activity page is not visible with BP 5.0.0 with Nouveau template._

@kjnanda / @pooja-muchandikar 
We have `Enable Direct Upload` Feature in rtmedia
So if this feature is enabled then the `Post Update` button should be visible or not?

kjnanda:
It should be visible but it will not need to be clicked on.

dishitpala:
@kjnanda I have created PR for
- Post update button on Activity page is not visible with BP 5.0.0 with Nouveau template.

While testing I faced some other CSS issues as well ( All buttons are not indented properly )
![Activity – admin – rtmedia (2)](https://user-images.githubusercontent.com/25063903/66629249-d1877400-ec1e-11e9-95b3-f4cd19f2fd50.png)

Above image testing environment 
```
BuddyPress: 3.0.0
Theme: TwentyNineteen
Rtmedia: 4.5.8
```
Please can you check this at your end

kjnanda:
@dishitpala We can skip CSS part as of now. `Post Update` should be there which is in priority.

@thrijith will that be fine?

kjnanda:
This is working as expected and tested with the `pre-release` branch. ✅

pooja-muchandikar:
@dharmin @thrijith 
Tested with the following test environment
```
WP: v5.2.4
BP: v5.0.0
rtMedia: v4.5.8 pre-release branch
```

Tested this scenario with the below pages: 
Sitewide Activity
User's activity
Group - Activity 

And it works fine.

pooja-muchandikar:
Tested with develop branch and its working fine. 
```
WP: v5.2.4
BP: v5.0.0
rtMedia: v4.5.9 develop branch
```

pooja-muchandikar:
Tested with the master branch and its working fine.

PR #1433, opened 2019-10-11
Fix Post Update button issue with BuddyPress 5.0.0

dishitpala:
#1432 _1. Post update button on Activity page is not visible with BP 5.0.0 with Nouveau template._

- Fix Post Update button issue with BuddyPress 5.0.0
- Tested with BuddyPress Version 3.0.0, 4.0.0 and 4.3.0
- Working fine with the above version's Nouveau template. and legacy template.


app/assets/js/rtMedia.backbone.js
```diff
@@ -1005,8 +1005,14 @@ jQuery( document ).ready( function( $ ) {
 
 			var rt_uploader_div = whats_new_form.find( '.rtmedia-uploader-div' );
 			var rt_uploader_filelist = whats_new_form.find( '#rtmedia_uploader_filelist' );
+			var whats_new_option = whats_new_form.find( '#whats-new-options' );
 
 			rt_uploader_div.show();
+			whats_new_option.show();
```
thrijith: @dishitpala Before performing actions, should we check length of the object to make sure it exits?

dishitpala: make sense 👍

PR #1438, opened 2019-10-24
Fix when attachments were not getting deleted on activity delete

dharmin:
Remove `is_admin()` condition and add a condition to check the `current_screen`.

### DIFF
--- a/app/main/controllers/media/RTMediaMedia.php
+++ b/app/main/controllers/media/RTMediaMedia.php
@@ -472,7 +472,7 @@ class RTMediaMedia {
 				}
 			}
 
-			if ( ! $core && ! is_admin() ) {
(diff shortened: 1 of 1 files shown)


kapilpaul:
@pooja-muchandikar please test this PR.

pooja-muchandikar:
@kapilpaul 

The changes done in this PR are reflecting, but since there is no issue mentioned in PR description not sure what was the issue. 

I checked with master branch and media are getting deleted as expected so exact what change is done not sure😕

pooja-muchandikar:
Hi @kapilpaul,

Got details about this issue. Issue was the media was not getting deleted even after deleting it from activity page.

- Upload media in activity page
- Open media in new tab and also open your media page
- Delete media from activity page
- Refresh the 2 tabs opened the media should get deleted

Its working as expected so you can merge this PR

PR #1453, opened 2019-11-13
Copy original "Post Update" button value

axettone:
Instead of re-creating the Post Update button, copy its original value! Why? 
Because sometimes it has been translated and in that case the user would see the button value change to "Post Update" after picking the media to upload.

### DIFF
--- a/app/assets/js/rtMedia.backbone.js
+++ b/app/assets/js/rtMedia.backbone.js
@@ -1887,7 +1887,7 @@ function rtmedia_selected_file_list( plupload, file, uploader, error, comment_me
 				class : 'button',
 				name  : 'aw-whats-new-submit',
 				id    : 'aw-whats-new-submit',
-				value : 'Post Update'
+				value : new_button.val()
 			} );
 
 			new_submit_btn.replaceWith( new_button );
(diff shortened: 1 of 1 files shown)


axettone:
Travis failed, but my single-line-change has nothing to do with that.

kapilpaul:
@pooja-muchandikar please test this PR

pooja-muchandikar:
Hi @kapilpaul 

I checked this PR and Post Update button is appearing same. There is no difference though. Can you check here?

Screenshot 2021-09-30 at 1 27 13 PM

kapilpaul:
@pooja-muchandikar please change your site language and then test again.

pooja-muchandikar:
@kapilpaul 

This PR is about these changes 👇 

Screenshot 2021-10-04 at 4 51 14 PM

kapilpaul:
yes. just change to master/develop and check what happens for this?

pooja-muchandikar:
@kapilpaul 

Got the issue but looks like with these changes unable to select media. Following error is shown
Screenshot 2021-10-04 at 5 14 57 PM

kapilpaul:
@pooja-muchandikar I have fixed that. please check again.

pooja-muchandikar:
Hi @kapilpaul 

Cross-checked the issue and the fix is working as expected now you can merge the PR.

PR #1461, opened 2019-11-19
Fix vulnerability in symfony/dependency-injection composer package

kjnanda:
1. Update codeception to the latest version (3.1.2)
2. Update deprecate function notice for bootstrap. It is fixed in codeception.yml file


kjnanda:
@chandrapatel Please merge this PR into pre-release

Issue #1449, opened 2019-11-08
Enable media in comment setting should have dependent check for Enable comment on uploaded media


kjnanda:
We have two different settings in rtmedia.

1. Display Tab - ```Allow user to comment on uploaded media``` 
2. BuddyPress Tab - ```Enable media in comment```

Now, the settings in the BuddyPress tab will not make sense if the setting in the Display tab is disabled. We should add this validation that if `Allow users to comment on uploaded media` is not enabled then `Enable media in comment` setting should get disabled automatically.

kjnanda:
Closing this issue w.r.t this comment  https://github.com/rtMediaWP/rtMedia/pull/1462#issuecomment-595689104

PR #1462, opened 2019-11-19
Add dependency on "Enable media in comment settings"

dishitpala:
#1449 
In rtMedia settings:
If `Allow user to comment on uploaded media` in `Display` tab is disabled
Then `Enable media in comment` in `BuddyPress` tab can not be enabled

`Enable media in comment` is dependent on `Allow user to comment on uploaded media`


kjnanda:
@thrijith This issue is reported by me only. However, I feel that this scenario is invalid. 

For rtMEdia it is valid but applying this workaround would also disable the media upload on comment for activity page. So we should not merge this PR and I am marking the issue as invalid.
CC @dishitpala

Issue #1463, opened 2019-11-20
Show count on tabs in Nouveau Template


kjnanda:
```
Test Environment
rtMedia: 4.5.9
BuddyPress: 5.0.0
Themes: Default theme
```
We can show count on the tabs in the nouveau template as its shown in the legacy template. 
1. Single Group page - Media count can be shown. 
Refer Legacy template screenshot wherein media count is shown: https://drive.google.com/open?id=1giGQB4WZsu0OEhW_7h2cWIApU9O7_DDO
Refer Nouveau template screenshot wherein media count is not shown: https://drive.google.com/open?id=14BhT_poH3iIL-t0Sw-vJ11gLGVJETb3V
##

2. Profile Activity page as well we can show count for Notification, Friends, Media etc in Nouveau template.
Refer  Legacy template screenshot wherein count is shown: https://drive.google.com/open?id=1W9QpFGhV8_XPLkdav7vCOVSGf7tNpuHL
Refer Nouveau template screenshot wherein count is not shown: https://drive.google.com/open?id=1x1nx4qBfo5JgRTXaKI2kH544v_SDTc49
##

pooja-muchandikar:
@SID177 

```
Test Environment
WordPress: v5.4-RC-2
rtMedia: v4.6.0  pre-release branch
BuddyPress: v5.1.2 Legacy and Nouveau
Theme: 2020 2019
```

Tested with `pre-release` branch and following are my findings . 

> Single Group page - Media count can be shown.

This is fixed :heavy_check_mark: 

---

> Profile Activity page as well we can show count for Notification, Friends, Media etc in Nouveau template.

This is not fixed :x:  For the above issue in Legacy template if there are no notifications, messages count is shown `0`. In Nouveau template the count `0` is not shown. But when there are notifications and messages the count is shown in Nouveau template. Which is correct. But since, we are trying to show consistent count to make legacy and nouveau template similar then if there are no notifications should `0` be shown in Nouveau template?

@kjnanda please share your thought.

kjnanda:
@pooja-muchandikar We should show the count as `0` when there is nothing in that specific tab.

pooja-muchandikar:
@SID177 

Please note the comment  https://github.com/rtMediaWP/rtMedia/issues/1463#issuecomment-597454975 and do the necessary changes.

In nouveau template of there are no notifications or messages count `0` should be shown.

SID177:
[...]

PR #1495, opened 2019-12-03
Fix Show count on tabs in Nouveau Template issue

dishitpala:
#1463 
Activity – admin – rtmedia


app/main/controllers/template/RTMediaNav.php
```diff
@@ -454,4 +456,95 @@ function set_privacy( $profile ) {
 
 		return $privacy;
 	}
+
+	/**
+	 * Check if media component has counts.
+	 *
+	 * @param bool   $count  Component feed has count.
```
thrijith:
```suggestion
	 * @param bool   $has_count  True if the nav has a count attribute. False otherwise
```
Could we use an other variable name than count, since the variable has boolean value, `count` as a name is misleading.

app/main/controllers/template/RTMediaNav.php
```diff
@@ -454,4 +456,95 @@ function set_privacy( $profile ) {
...
+		if ( 'personal' === $value ) {
+			$media = $this->actual_counts( $bp->displayed_user->id );
+		}
+
+		if ( empty( $media ) ) {
+			return $count;
+		}
+
+		// Convert to number.
+		$media_count = rtmedia_number_to_human_readable( $media['total']['all'] );
```
thrijith: Should empty check be added for `$media['total']` and `$media['total']['all']`, also since the function expects the param to be int may be use `absint()`

PR #1500, opened 2019-12-09
Phpcs resolved in controllers API, group and privacy folders

vaishaliagola27:


app/main/controllers/privacy/RTMediaPrivacy.php
```diff
@@ -220,32 +267,44 @@ function select_privacy_ui( $echo = true, $element_id = false, $element_class =
 		}
 
 		if ( $echo ) {
-			echo $form->get_select( $attributes ); // @codingStandardsIgnoreLine
+			echo wp_kses( $form->get_select( $attributes ), wp_kses_allowed_html() );
```
chandrapatel: @vaishaliagola27 Make sure this won't break the output.

vaishaliagola27: Sure. Will check before merging.

PR #1504, opened 2019-12-13
phpcs resolved in templates folder files 

vaishaliagola27:


templates/main.php
```diff
@@ -1,50 +1,53 @@
 <?php
-/* * **************************************
+/**
+ * **************************************
```
chandrapatel: Remove `**************************************`

templates/main.php
```diff
@@ -1,50 +1,53 @@
...
  * Main.php
  *
  * The main template file, that loads the header, footer and sidebar
  * apart from loading the appropriate rtMedia template
+ *
+ * @package rtMedia
  * *************************************** */
```
chandrapatel: Remove `**************************************`

PR #1544, opened 2020-02-04
Fix GH-1535: Separated image and text using DOMDocument class

### LINKED ISSUE #1535: On activity page show long text with read more along with the image
```
Test Environment
WordPress: v5.3.2
BuddyPress: v5.1.2
rtMedia: v4.5.11
```

**Steps to Reproduce:**
1. Go to activity page
2. Enter long text along with the image
3. Observe that if long text is entered then read more is shown but the image is shown only after clicking on read more. 
We need to show the long text truncated with a read more option and the image displayed below the truncated text.

Refer to the Visual -> https://drive.google.com/file/d/1ARhvuXPYg3LteBn0ERqLCeL7nViPolK8/view

SID177:
I used DOM manipulation class of php (DOMDocument) to separate image from text. I created custom excerpt where I only trimmed text and kept other HTML as it is.

Fixes #1535

### DIFF
--- a/app/main/controllers/activity/RTMediaBuddyPressActivity.php
+++ b/app/main/controllers/activity/RTMediaBuddyPressActivity.php
@@ -96,6 +96,11 @@ class RTMediaBuddyPressActivity {
 		// So we need to iterate.
 		$div_list = $dom->getElementsByTagName( 'div' );
 
+		// Return if no divs found.
+		if ( empty( $div_list ) ) {
+			return $excerpt;
+		}
+
 		// We're storing first div to create final markup.
 		// If we create markup from dom object, it'll create whole HTML which we don't want.
 		$first_div = '';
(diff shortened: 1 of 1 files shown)


umeshnevase:
@SID177 We need to provide settings to enable disable this feature.

app/main/controllers/activity/RTMediaBuddyPressActivity.php
```diff
@@ -64,6 +67,81 @@ function __construct() {
...
+		$dom->loadHTML( $text );
+		// We need to find div having rtmedia-activity-text class, but no direct method for it.
+		// So we need to iterate.
+		$div_list = $dom->getElementsByTagName( 'div' );
+
+		// We're storing first div to create final markup.
+		// If we create markup from dom object, it'll create whole HTML which we don't want.
+		$first_div = '';
+
+		foreach ( $div_list as $div ) {
```
umeshnevase: Check empty $dev_list

app/main/controllers/activity/RTMediaBuddyPressActivity.php
```diff
@@ -64,6 +67,81 @@ function __construct() {
...
+			}
+
+			// We need div with class attribute.
+			if ( empty( $div->attributes ) ) {
+				continue;
+			}
+
+			$atts = $div->attributes;
+			// Check attributes by iterating them.
+			foreach ( $atts as $att ) {
```
umeshnevase: Check !empty($atts)

SID177: This is already checked on line 114 `if ( empty( $div->attributes ) ) {`

PR #1568, opened 2020-03-03
Fix GH-1567: Add utf-8 encoding in DOMDocument loadHTML method

### LINKED ISSUE #1567: Different languages not showing properly
We need to add `<?xml encoding="utf-8" ?>` while using `DOMDocument` class.

Without adding utf-8 encoding.
Screenshot 2020-03-03 at 5 04 21 PM

With utf-8 encoding
Screenshot 2020-03-03 at 5 05 46 PM

SID177:
Prepend `<?xml encoding="utf-8" ?>` with activity text in loadHTML method of DOMDocument's class.
Fix #1567

### DIFF
--- a/app/main/controllers/activity/RTMediaBuddyPressActivity.php
+++ b/app/main/controllers/activity/RTMediaBuddyPressActivity.php
@@ -108,7 +108,8 @@ class RTMediaBuddyPressActivity {
 		$dom = new DOMDocument();
 		// DOMDocument gives error on html5 tags, so we need to disable errors.
 		libxml_use_internal_errors( true );
-		$dom->loadHTML( $text );
+		// Use utf-8 encoding for different languages.
+		$dom->loadHTML( '<?xml encoding="utf-8" ?>' . $text );
 		// DOMDocument gives error on html5 tags, so we need to disable errors.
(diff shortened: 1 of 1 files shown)


app/main/controllers/activity/RTMediaBuddyPressActivity.php
```diff
@@ -108,7 +108,8 @@ public function bp_activity_truncate_entry( $excerpt, $text, $readmore ) {
 		$dom = new DOMDocument();
 		// DOMDocument gives error on html5 tags, so we need to disable errors.
 		libxml_use_internal_errors( true );
-		$dom->loadHTML( $text );
+		// Use utf-8 encoding for different languages.
+		$dom->loadHTML( '<?xml encoding="utf-8" ?>' . $text );
```
abhijitrakas:
```suggestion
		$dom->loadHTML( '<?xml encoding="utf-8" ?>'  );
		$dom->loadHTML($text);
```
Can we write it like ☝️

SID177: Doesn't work as expected. I think we can loadHTML only once.

PR #1571, opened 2020-03-09
Fix #1565: Remove duplicate IDs

### LINKED ISSUE #1565: [Customer Query] DOM Element warnings on activity page
DOM element warnings are shown on activity page.

Screenshot from 2020-03-02 13-37-29

----

Update this customer once release it out - https://wordpress.org/support/topic/non-unique-id-rt_upload_hf_comment/

SID177:
Remove IDs from elements rendering multiple times on same page.
Fixes #1565


app/main/controllers/upload/RTMediaUploadView.php
```diff
@@ -274,6 +278,15 @@ public function render( $template_name ) {
 				if ( ( 'upload_parent_id' === $key && ! empty( $key ) ) || ( 'upload_parent_id_type' === $key && ! empty( $key ) ) ) {
 					$selector = 'class';
 				}
+
+				/**
+				 * These will be repeated for all activities, which will result in duplicate elements with same IDs.
+				 * So adding these as class.
+				 */
+				if ( 'comment' === $key || 'privacy' === $key || 'upload_parent_id_context' === $key ) {
```
abhijitrakas: can we use `in_array` here?

app/main/controllers/media/RTMediaLike.php
```diff
@@ -390,7 +389,8 @@ public function before_render() {
 	 * @return string
 	 */
 	public function like_button_filter_nonce( $button ) {
-		$button .= wp_nonce_field( 'rtm_media_like_nonce' . $this->media->id, 'rtm_media_like_nonce', true, false );
+		// This nonce is rendered multiple times on same page, so having common ID will result in HTML warning.
+		$button .= '<input type="hidden" name="rtm_media_like_nonce" value="' . esc_attr( wp_create_nonce( 'rtm_media_like_nonce' . $this->media->id ) ) . '" />';
```
abhijitrakas: Here instead of adding custom HTML can we update ID?

SID177: There are 2 like buttons for same media, so even if we add ID with media_id appended it'll be the same ID.

abhijitrakas: We can load this nonce one as it is common for both. Means just load once instead of two times. You can add check if it already loaded then don't load the second time. This way no need to add a custom HTML field.

SID177: Yes we can do it. But is it necessary to give ID? `var like_nonce = $( this ).siblings( 'input[name="rtm_media_like_nonce"]' ).val();` this is being searched in siblings so the performance won't be affected much.

SID177: And even if the nonce is common, the form for both like buttons is different. If we add common nonce with ID, we'll have to fetch that nonce using JS on form submit (like `var like_nonce = $( this ).siblings( 'ID' ).val();`). In this case if JS is disabled it won't work.

abhijitrakas: Are we using form serialize to submit data?

SID177: Not form serialization, we're just sending nonce in the form. https://github.com/rtMediaWP/rtMedia/pull/1571/files#diff-178843aed6652ee542e53006028e674aR1786
But if JS is enabled it's handled by ajax otherwise I think it'll be submitted

abhijitrakas: Nope. I don't think for data will submit if JS is disabled. Please check the same scenario and if not happening something like that then we could use `wp_nonce_field` this function once only to load a single field.

SID177: Form data will be submitted when JS is disabled but won't be able to verify nonce because we're checking for `like_nonce` which is being set in JS only.
So I think the change to make here is to change the nonce name to `like_nonce` from `rtm_media_like_nonce`. But I don't see any way to add only 1 nonce and make it work when JS is disabled.

app/main/controllers/media/RTMediaLike.php
```diff
@@ -11,11 +11,20 @@
  */
 class RTMediaLike extends RTMediaUserInteraction {
 
+	/**
+	 * Check whether like nonce is already added or not.
+	 *
+	 * @var boolean
+	 */
+	private $like_nonce_loaded;
```
abhijitrakas:
```suggestion
	private $like_nonce_loaded = false;
```

app/main/controllers/media/RTMediaLike.php
```diff
@@ -11,11 +11,20 @@
...
+
 	/**
 	 * RTMediaLike constructor.
 	 */
 	public function __construct() {
 
+		$this->like_nonce_loaded = false;
```
abhijitrakas:
```suggestion		
```

app/main/controllers/upload/RTMediaUploadView.php
```diff
@@ -38,7 +38,11 @@ public function __construct( $attr ) {
 	public static function upload_nonce_generator( $echo = true, $only_nonce = false ) {
 
 		if ( $echo ) {
-			wp_nonce_field( 'rtmedia_upload_nonce', 'rtmedia_upload_nonce' );
+			/**
+			 * If we use wp_nonce_field, it'll create a hidden field with ID, which will generate HTML warning.
+			 * Because this nonce is being used in multiple places on the same page.
+			 */
+			echo '<input type="hidden" name="rtmedia_upload_nonce" value="' . esc_attr( wp_create_nonce( 'rtmedia_upload_nonce' ) ) . '" />';
```
abhijitrakas: Do we need this now?

SID177: The ajax upload of activity comment is handled in BuddyPress. So we'll have to keep individual nonce for each upload form.

PR #1574, opened 2020-03-11
Code refactor

kjnanda:
1. Remove unwanted code from composer.json file
2. Update constant value for customcss var in Constants.php file
3. Remove unwanted hardcoded wait() from Login.php file
4. Update code in EnableRtmediaDefaultStlyeWithCustomCssCept.php file


kjnanda:
@SID177 You can merge this PR.

@juhi123 As per our discussion, We are going to skip the test case under folder `06`.

However, I could not find a way to skip the cept file. There are ways to skip the test and cest files using the code or commands. 
So as a workaround I did this to avoid the failure of test cases - https://travis-ci.org/github/rtMediaWP/rtMedia/builds/661393991#L3727

Used `assertNotEqual()`

PR #1576, opened 2020-03-11
Update custom css test case and travis build

juhi123:


kjnanda:
@juhi123 We can close this PR, right?

PR #1582, opened 2020-03-16
Moderation issue, i.e., media privacy changes after allowing moderated media from admin

vermadarsh:
This PR relates to the issue in the `Moderation` add-on. Link: https://github.com/rtCamp/rtmedia-moderation/issues/26

I've added the code check against `false` just for the reason when the saved privacy value is 0 and in the `if` condition, becomes false, that's why added strict type-checking.

### DIFF
--- a/app/main/controllers/media/RTMediaMeta.php
+++ b/app/main/controllers/media/RTMediaMeta.php
@@ -72,7 +72,7 @@ class RTMediaMeta {
 		if ( true === $duplicate ) {
 			$media_meta = $this->model->insert( array( 'media_id' => $id, 'meta_key' => $key, 'meta_value' => $value ) );
 		} else {
-			if ( $this->get_single_meta( $id, $key ) ) {
+			if ( false !== $this->get_single_meta( $id, $key ) ) {
 				$meta       = array( 'meta_value' => $value );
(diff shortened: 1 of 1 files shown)


app/main/controllers/media/RTMediaMeta.php
```diff
@@ -72,7 +72,7 @@ public function update_meta( $id = false, $key = false, $value = false, $duplica
 		if ( true === $duplicate ) {
 			$media_meta = $this->model->insert( array( 'media_id' => $id, 'meta_key' => $key, 'meta_value' => $value ) );
 		} else {
-			if ( $this->get_single_meta( $id, $key ) ) {
+			if ( false !== $this->get_single_meta( $id, $key ) ) {
```
SID177: So for privacy `logged-in-users`, the meta value is empty?

vermadarsh: No, the privacy for `public` in 0.
When we change it to something else, it goes for checking, if a meta entry already exists. So, for `public` privacy, the entry exists which returns 0 as the meta value.
Now, this 0 is checked against `if` condition (for insert or updating the entry)
When we check 0 inside `if`, i.e., `if( 0 )`
This results in a falsy condition check which creates another row for `logged-in-users` privacy moderation.
Ideally, it should update the same entry.

Issue #1584, opened 2020-03-19
Issue with default privacy 


kjnanda:
[pre-release branch] The Privacy tab does not show default selected value once the plugin is installed. Also, it does not show the selected state does not remain after clicking on the **Save Settings** button

-------

![Settings_‹_beta_release_testing_—_WordPress](https://user-images.githubusercontent.com/15173772/77077617-781c4000-6a1b-11ea-8987-593d99444d8c.png)

kjnanda:
This has been fixed with above PRs

PR #1585, opened 2020-03-20
Fixed the issue, radio button bot showing as selected after saving admin settings

vermadarsh:
This PR pertains to issue #1584
The radio button does not show to be checked even after saving admin settings, which is fixed.

Screenshot attached.
default-privacy-issue-fixed


app/admin/RTMediaFormHandler.php
```diff
@@ -143,7 +143,7 @@ public static function checkbox( $args, $echo = true ) {
 	 *
 	 * @access static
 	 *
-	 * @param  array $args arguments to show radio buttons.
+	 * @param  array $args Arguments to show radio buttons.
```
SID177:
```suggestion
	 * @param array $args Arguments to show radio buttons.
```

vermadarsh: Extra space removed.

PR #1588, opened 2020-03-20
Fixed the error of radio buttons, not working after saving settings.

vermadarsh:
This PR pertains to issue #1584 

This PR fixes the `checked` state of the radio buttons in plugin admin settings. The radio buttons after being checked were not showing as checked after saving settings.

Screenshots:
Privacy Settings: https://prnt.sc/rjj4c4
List Media View: https://prnt.sc/rjj4qc

### DIFF
--- a/app/admin/RTMediaFormHandler.php
+++ b/app/admin/RTMediaFormHandler.php
@@ -170,7 +170,7 @@ class RTMediaFormHandler {
 		foreach ( $args['radios'] as $value => $key ) {
 			$args['rtForm_options'][] = array(
 				$key      => $value,
-				'checked' => ( strval( $args['default'] ) === strval( $value ) ) ? true : false, // strval is used to fix the privacy levels being integer to match the saved privacy value.
+				'checked' => ( strval( $args['default'] ) === strval( $value ) ) ? true : false, // The function 'strval()' is used to fix the privacy levels being integer to match the saved privacy value.
 			);
 		}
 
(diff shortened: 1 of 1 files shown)


app/admin/RTMediaFormHandler.php
```diff
@@ -170,7 +170,7 @@ public static function radio( $args ) {
 		foreach ( $args['radios'] as $value => $key ) {
 			$args['rtForm_options'][] = array(
 				$key      => $value,
-				'checked' => ( (int) $args['default'] === $value ) ? true : false,
+				'checked' => ( $args['default'] == $value ) ? true : false, // phpcs:ignore WordPress.PHP.StrictComparisons.LooseComparison The == error ignored as the privacy levels indexes are integer, and while saving in database, the saved value becomes string.
```
SID177:
```suggestion
				'checked' => ( $args['default'] == $value ) ? true : false, // phpcs:ignore WordPress.PHP.StrictComparisons.LooseComparison -- The == error ignored as the privacy levels indexes are integer, and while saving in database, the saved value becomes string.
```

Did you test this by running PHPCS? It'll give error for loose comparison in your current code. `// phpcs:ignore` won't work like this. Please test before committing the code.

vermadarsh: Added `--` in phpcs ignore comment.

app/admin/RTMediaFormHandler.php
```diff
@@ -170,7 +170,7 @@ public static function radio( $args ) {
 		foreach ( $args['radios'] as $value => $key ) {
 			$args['rtForm_options'][] = array(
 				$key      => $value,
-				'checked' => ( (int) $args['default'] === $value ) ? true : false,
+				'checked' => ( strval( $args['default'] ) === strval( $value ) ) ? true : false, // The function 'strval()' is used to fix the privacy levels being integer to match the saved privacy value.
```
SID177: Keep the comment above the code.

vermadarsh: I moved the comment above the code.

Issue #1593, opened 2020-03-25
Round of testing with pre-release branch of rtMedia core with other add-ons


kjnanda:
- [x] **1. Facing the issue while posting the media/doc in bbPress forum. I am able to upload the media but first, it takes me to this below page and then going back it shows that media is posted and also, uploaded media is not removed from the uploader panel.**

----

![localhost_beta-testing_groups_test-grp_forum_topic_first-topic__post-187](https://user-images.githubusercontent.com/15173772/77514743-5e587e00-6e9d-11ea-877a-bd61575906d0.png)

-----

![Forum_–_Test_grp_–_beta_release_testing](https://user-images.githubusercontent.com/15173772/77514945-ba230700-6e9d-11ea-8e5b-aa34e9a607d3.png)

-----------

- [x] **2. Custom thumbnail add-on functionality is broken**

Please refer to this visual where I have shown how this admin-related setting does not work for custom thumbnail add-on w.r.t pre-release branch.

https://drive.google.com/file/d/1NGF5wcwnAaJ7b9p26lqyORYFKHAiYxvP/view

- [x] **3. Upload by Url add-on. When terms of services are enabled along with direct upload, it bypasses the validation and the file get uploaded without checking terms of services checkbox.**

https://drive.google.com/file/d/1W0N4mDPVFQHzGNqgAyf_AK35owS5sWgP/view

-------

- [x] **4. rtmedia Wordpress comment attachment** - Facing the same issue as mentioned in subtask 1

Please do the needful.

SID177:
>4. rtmedia Wordpress comment attachment - Facing the same issue as mentioned in subtask 1

@kjnanda This seems to be working fine for me for `pre-release` of rtMedia core.
![Hello world! – localfw-rtmedia test](https://user-images.githubusercontent.com/23107546/77902693-3c953780-729f-11ea-86f0-60eebd85a0b5.gif)

kjnanda:
@SID177 Please check my comment below.

Subtask - 1 ✅ 
Subtask 2 - Seems to be pending
Subtask 3 - Code needs to be merged into the branch
Subtask - 4 Works fine now. 

The reason is, you added the fix for subtask 1 and it was a similar issue hence you could not replicate it after the added fix.

kjnanda:
@SID177 Added changes LGTM 👍

kjnanda:
@SID177 Please check the below condition where upload operation is not working.

[...]

PR #1597, opened 2020-04-01
Fix photos not uploading for custom thumbnail addon

SID177:
This issue was caused by PHPCS fixes.
Fixes: 2nd of https://github.com/rtMediaWP/rtMedia/issues/1593

### DIFF
--- a/app/helper/RTMediaSettings.php
+++ b/app/helper/RTMediaSettings.php
@@ -233,6 +233,11 @@ if ( ! class_exists( 'RTMediaSettings' ) ) {
 
 				$http_referer = rtm_get_server_var( 'HTTP_REFERER', 'FILTER_SANITIZE_URL' );
 				if ( isset( $http_referer ) ) {
+					/**
+					 * We're not adding exit() here because the functionality is breaking.
+					 * I tried moving this code at the end of the function but it's still not working,
+					 * maybe there is another function/code which is being executed after this function.
+					 */
 					wp_safe_redirect( $http_referer . $settings_saved );
 				}
 				global $rtmedia;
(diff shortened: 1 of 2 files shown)


app/helper/RTMediaSettings.php
```diff
@@ -234,7 +234,6 @@ public function settings() {
 				$http_referer = rtm_get_server_var( 'HTTP_REFERER', 'FILTER_SANITIZE_URL' );
 				if ( isset( $http_referer ) ) {
 					wp_safe_redirect( $http_referer . $settings_saved );
-					exit();
```
ravatparmar, line 237: Note: wp_safe_redirect() does not exit automatically, and should almost always be followed by a call to exit; 

Can we check if it's any other issue because wp_safe_redirect needs to have exit, or we can put exit conditionally if it's successfully redirected.

SID177: @ravatparmar I understand what you're saying, I read the doc. I tried many things to add `exit()` but it just doesn't work that way. There is a code running after this function so I can't add `exit()` in this function. So I think the code was written this way.
`exit()` was added as phpcs fix, before adding it was working fine.
I think we should keep it this way because this function is common for all rtMedia settings. Making change here might break other things.

ravatparmar: As you said here `There is a code running after this function`, if possible can we move this redirection at end of the function? so code will be executed as normal and then it'll redirected to the next specific page.

SID177: I already tried moving it at the end of the function, not working.

ravatparmar: Okay, one last thing, before `wp_safe_redirect( $http_referer . $settings_saved );`, can you please write down comment why we are not putting exit after `wp_safe_redirect` with the issue so later it can be remembered and don't get this issue again.

SID177: Added the comment

Issue #1477, opened 2019-11-21
Incorrect css class on the error message of media update.


kjnanda:
HTML in media_update_success_error() action have incorrect css class rtmedia-error instead of rtmedia-warning.

PR: https://github.com/rtMediaWP/rtMedia/pull/1611

kjnanda:
@SID177 What output am I supposed to expect here? Can you provide me a visual of before fix and after fix?

SID177:
@kjnanda 
Before adding fix
![Screenshot 2020-04-20 at 11 29 35 AM](https://user-images.githubusercontent.com/23107546/79718744-6b308c00-82fa-11ea-86ed-8fa29d66513b.png)

After adding fix
![Screenshot 2020-04-20 at 11 30 12 AM](https://user-images.githubusercontent.com/23107546/79718757-708dd680-82fa-11ea-8fea-b5d7f236db1b.png)

kjnanda:
@SID177 Please check the below screenshot.

![Media_–_ADMINUSER_–_beta_release_testing](https://user-images.githubusercontent.com/15173772/79748372-f1fd5d00-832a-11ea-8d70-cd613a7ebb2f.png)

I am seeing this message twice upon saving the media after changes. 

Also, I am not able to generate the error you have shown in the screenshot shared by you!

kjnanda:
@SID177 I am not able to understand this issue. 

With this code `if ( false == $state ) {` I see the below screenshot! 

----

![Media_–_ADMINUSER_–_beta_release_testing](https://user-images.githubusercontent.com/15173772/79843384-d3a16b00-83d7-11ea-8041-faef7e652b43.png)

---
It will be really great if you can further debug this issue and can point me in the right direction.

SID177:
@kjnanda I've fixed the media update message showing twice issue in PR - https://github.com/rtMediaWP/rtMedia/pull/1620
The code is merged in `pre-release`

SID177:
@kjnanda Make sure you have latest `pre-release` branch. Try hard refreshing the browser. If it doesn't work then inspect the error message and check whether the div has `rtmedia-warning media-edit-messge` class on it.
If it has `rtmedia-error media-edit-messge` class on then I think you don't have the latest code.

Check `app/main/controllers/template/RTMediaTemplate.php` file's 475 line. The div should have `rtmedia-warning media-edit-messge` class if it's latest code.

kjnanda:
@SID177 This is working fine 👍

[...]

PR #1611, opened 2020-04-17
Fixes - 1477: Add filter docs, change css class of error message div, use sprintf

SID177:
- Added documentation comment for filters. I couldn't find any usage of last bool parameter but I'm not removing it in case any theme/user is using it.
- Changed error message div class
- Used `sprintf`

Fixes https://github.com/rtMediaWP/rtMedia/issues/1477

### DIFF
--- a/app/main/controllers/template/RTMediaTemplate.php
+++ b/app/main/controllers/template/RTMediaTemplate.php
@@ -459,12 +459,33 @@ class RTMediaTemplate {
 
 	/**
 	 * Show error message for media update error.
+	 *
+	 * @return void
 	 */
 	public function media_update_success_error() {
-		$message = apply_filters( 'rtmedia_update_media_message', esc_html__( 'Error in updating Media', 'buddypress-media' ), true );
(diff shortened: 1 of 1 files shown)


app/main/controllers/template/RTMediaTemplate.php
```diff
@@ -459,12 +459,33 @@ public function media_update_success_messege() {
...
+	 * @return void
 	 */
 	public function media_update_success_error() {
-		$message = apply_filters( 'rtmedia_update_media_message', esc_html__( 'Error in updating Media', 'buddypress-media' ), true );
-		$html    = "<div class='rtmedia-error  media-edit-messge'>" . esc_html( $message ) . '</div>';
+		/**
+		 * To filter error message on media update.
+		 *
+		 * @param string $message Raw error message.
+		 * @param bool   $bool
```
ravatparmar, line 470: Can we have description for bool parameter as well, here and bottom as well.

SID177: @ravatparmar Please check PR description, there I've explained this.

Issue #1608, opened 2020-04-16
[Uncaught Error:] Fix call to undefined function error thrown on media page when nouveau template is active and BuddyPress is inactive


kjnanda:
### Issue Description

When the Nouveau template is active then the media page shows many errors.

----

### Steps to reproduce an issue

1. Activate rtMedia and BuddyPress both
2. Enable Nouveau template
3. Goto /media page in the front end
4. Now, deactivate BuddyPress
5. Refresh the media page 
6. Observe that it throws many errors

----

### Attachment

![Media_–_test](https://user-images.githubusercontent.com/15173772/79445219-e98ae680-7ff9-11ea-9b73-1e1130f27fd8.png)

----

### Test Environment Used

```
WordPress - V5.4
BuddyPress - V5.1.2
rtMedia - V4.6.1
```
-----

PS: One of our customers also faced the same issue. `2572585` Please update him once the issue fixed and released.

kjnanda:
@SID177 Please check my comment below.

1. When Legacy was active before BuddyPress gets deactivated then it shows proper the media available on the page

![Media_–_test](https://user-images.githubusercontent.com/15173772/79562184-3f2dc480-80c8-11ea-904d-27169176879d.png)

-----

2. When Nouveau was active before BuddyPress gets deactivated then it does not show the media available on the page

![Media_–_test](https://user-images.githubusercontent.com/15173772/79562400-a9df0000-80c8-11ea-9248-170877df9e35.png)

SID177:
@kjnanda Are we supposed to handle showing `/media` page when BuddyPress is inactive?

kjnanda:
@SID177 Can you please elaborate on your query?

SID177:
@kjnanda I've showed the rtMedia template when Nouveau was last active, like it shows in case of Legacy template.

kjnanda:
@SID177 This looks better now.

kjnanda:
This is working as expected and tested with `dev` branch. ✅

kjnanda:
Closing this issue as this has been fixed and release in - V4.6.2

PR #1618, opened 2020-04-20
Fix 1608: Show rtmedia template in nouveau template like it's showing in legacy

SID177:
Show rtmedia template in nouveau template like it's showing in legacy when BuddyPress is inactive.
Fixes https://github.com/rtMediaWP/rtMedia/issues/1608#issuecomment-615184024

### DIFF
--- a/templates/main.php
+++ b/templates/main.php
@@ -166,32 +166,28 @@ if ( ! $rt_ajax_request ) {
 				}
 			}
 		} else { // if BuddyPress.
-			?>
-			<div id="item-body">
-			<?php
+			echo '<div id="item-body">';
 		}
 
 		rtmedia_load_template();
 
 		if ( ! $rt_ajax_request ) {
-			if ( function_exists( 'bp_displayed_user_id' ) && 'buddypress' === $template_type && ( bp_displayed_user_id() || bp_is_group() ) ) {
-				if ( bp_is_group() ) {
+			if ( 'buddypress' === $template_type ) {
+				if ( function_exists( 'bp_is_group' ) && bp_is_group() ) {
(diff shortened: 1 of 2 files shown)


templates/main.php
```diff
@@ -165,6 +165,36 @@
 					endwhile;
 				}
 			}
+		} else { // if BuddyPress.
+			?>
+			<div id="item-body">
```
ravatparmar, line 170: we can use printf if there is only one line, to make code more readable.

templates/main.php
```diff
@@ -165,6 +165,36 @@
...
+			<?php
+		}
+
+		rtmedia_load_template();
+
+		if ( ! $rt_ajax_request ) {
+			if ( function_exists( 'bp_displayed_user_id' ) && 'buddypress' === $template_type && ( bp_displayed_user_id() || bp_is_group() ) ) {
```
ravatparmar, line 177: I think here no need to put condition for this two functions: `bp_displayed_user_id() || bp_is_group()` as they are already getting checked in below if condition.

templates/main.php
```diff
@@ -165,6 +165,36 @@
...
+				}
+				if ( bp_displayed_user_id() ) {
+					do_action( 'bp_after_member_media' );
+					do_action( 'bp_after_member_body' );
+				}
+			}
+			?>
+			</div><!--#item-body-->
+			<?php
+			if ( function_exists( 'bp_displayed_user_id' ) && 'buddypress' === $template_type && ( bp_displayed_user_id() || bp_is_group() ) ) {
```
ravatparmar, line 190: here same as above: we are checking condition twice.

Issue #1624, opened 2020-04-22
Privacy Page redirection issue


kjnanda:
### Issue Description

Privacy settings show under the User's profile is no more accessible.

----

### Steps to reproduce an issue

Goto this link - http://<Site_Url>/members/<User_Name>/settings/privacy/

It should show a privacy option as shown in the below screenshot.

![Privacy_–_Settings_–_krupa_–_krupa_dev5_rt_gw](https://user-images.githubusercontent.com/15173772/79938656-d7d29480-847a-11ea-8d85-cc8c42b5fe8a.png)

----

### Attachment

![venting_site](https://user-images.githubusercontent.com/15173772/79938697-eb7dfb00-847a-11ea-80b5-2c1fab676c97.png)

----

### Test Environment Used

```
WordPress - V5.4
BuddyPress - V5.1.2
rtMedia - V4.6.1
```
----

`2574979`

kjnanda:
@SID177 Looks good now! 

1. The privacy page is visible now
2. Changing the privacy works fine
3. Disabling the privacy removes `privacy` option from settings
4. Viewing the privacy setting in logged out mode is not a valid scenario

Let me know if there is anything else we need to check.

SID177:
Changing privacy from logged out mode shouldn't be allowed and if such a thing happens then appropriate message should be displayed.
That's all I guess.

kjnanda:
@SID177 I found one weird scenario

If you try to access this link `http://localhost/beta-testing/members/adminuser/settings/` then it will redirect the user to log in page which is fair enough.

But if you try to access this `http://localhost/beta-testing/members/adminuser/settings/privacy/` it will allow selecting the radio button. However, changes are not happening and it says `No changes were made to your account.` but this also should not be allowed. 

IMO, it should have the same behavior as other tabs inside `settings`

SID177:
Yes, privacy page shouldn't be visible to users without login. But this behavior exists in older versions as well.
We can add a check in next version.

kjnanda:
Yes, We are good to freeze this release now.

kjnanda:
This is working as expected and tested with `dev` branch. ✅

kjnanda:
Closing this issue as this has been fixed and release in - V4.6.2

PR #1625, opened 2020-04-22
Fix GH-1624: Avoid redirection if redirect url matches current page url in privacy

SID177:
Avoid redirection if redirect url matches current page url in privacy
Fixes https://github.com/rtMediaWP/rtMedia/issues/1624

### DIFF
--- a/app/main/controllers/privacy/RTMediaPrivacy.php
+++ b/app/main/controllers/privacy/RTMediaPrivacy.php
@@ -487,7 +487,15 @@ class RTMediaPrivacy {
 
 			bp_core_add_message( $feedback, $feedback_type );
 			do_action( 'bp_core_general_settings_after_save' );
-			bp_core_redirect( bp_displayed_user_domain() . bp_get_settings_slug() . '/privacy/' );
+
+			// Function exists checks.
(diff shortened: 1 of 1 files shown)


app/main/controllers/privacy/RTMediaPrivacy.php
```diff
@@ -487,7 +487,15 @@ public function rt_privacy_settings_action() {
 
 			bp_core_add_message( $feedback, $feedback_type );
 			do_action( 'bp_core_general_settings_after_save' );
-			bp_core_redirect( bp_displayed_user_domain() . bp_get_settings_slug() . '/privacy/' );
+
+			// Function exists checks.
+			if ( function_exists( 'bp_get_requested_url' ) && function_exists( 'bp_displayed_user_domain' ) && function_exists( 'bp_get_settings_slug' ) ) {
```
ravatparmar, line 492: Just for information, if BuddyPress is installed then these are not their default function? can we check only one thing that if BuddyPress is there then this functions will be there, of only BuddyPress version checking.

SID177: I think it's best to use `function_exists` check before using any 3rd party functions. If they change or remove a function, atleast our code will never give fatal error for undefined function.

PR #1627, opened 2020-04-22
Fix GH-1624: Check default privacy string length instead of empty

SID177:
Check default privacy string length instead of empty.
Condition `intval( $default_privacy )` will always 0 which shouldn't happen and cause redirection loop.

Fixes regression issue caused by previous PR, issue - https://github.com/rtMediaWP/rtMedia/issues/1624

### DIFF
--- a/app/main/controllers/privacy/RTMediaPrivacy.php
+++ b/app/main/controllers/privacy/RTMediaPrivacy.php
@@ -466,8 +466,10 @@ class RTMediaPrivacy {
 		$default_privacy = sanitize_text_field( filter_input( INPUT_POST, 'rtmedia-default-privacy', FILTER_SANITIZE_STRING ) );
 		$nonce           = sanitize_text_field( filter_input( INPUT_POST, 'rtmedia_member_settings_privacy', FILTER_SANITIZE_STRING ) );
 
-		if ( ! empty( $default_privacy ) || 0 === intval( $default_privacy ) ) {
-
+		// Old condition won't work as we've added sanitize_text_field for $default_privacy.
+		// We can't perform empty as 0 could be the possible value, so we check for empty string instead.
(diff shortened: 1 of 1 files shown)


app/main/controllers/privacy/RTMediaPrivacy.php
```diff
@@ -466,8 +466,10 @@ public function rt_privacy_settings_action() {
 		$default_privacy = sanitize_text_field( filter_input( INPUT_POST, 'rtmedia-default-privacy', FILTER_SANITIZE_STRING ) );
 		$nonce           = sanitize_text_field( filter_input( INPUT_POST, 'rtmedia_member_settings_privacy', FILTER_SANITIZE_STRING ) );
 
-		if ( ! empty( $default_privacy ) || 0 === intval( $default_privacy ) ) {
-
+		// Old condition won't work as we've added sanitize_text_field for $default_privacy.
+		// We can't perform empty as 0 could be the possible value, so we check for empty string instead.
+		// Condition intval( $default_privacy ) will always 0 which shouldn't happen.
+		if ( 0 !== strlen( strval( $default_privacy ) ) ) {
```
ravatparmar, line 472: we can also use sanitize numeric, if that is the issue. what will be type of $default_privacy?

SID177: Converting `$default_privacy` to int is returning `0` so that won't work. We need to differentiate between 0 and empty.

ravatparmar: We can also check typeof , also it it's string we can put '' !==.

SID177: >also it it's string we can put '' !==.

That's what I've done. I haven't put `!==` instead I've validated it in string and checked it's length. So if it's empty (except 0 and other values) it won't go inside.

Issue #1633, opened 2020-04-27
Issue with Social sync add-on with rtMedi4.6.2 release


kjnanda:
Shows JavaScript lines of code on the site when `rtmedia social sync` add-on is active

-----

![screenshot-krupa dev5 rt gw-2020 04 27-11_45_04](https://user-images.githubusercontent.com/15173772/80339599-a7289b80-887c-11ea-897b-457382ae3ab4.png)

SID177:
@kjnanda Do I have to insert app details to see this issue? Otherwise I'm not getting this

kjnanda:
Yes, @SID177 

If you need you can use my site. You already have the credentials of my site. 

This is strange that the issue was not produced during the testing as I already conducted a test with all the add-ons! However, it appeared today. So let us fix this first. This has not been reported by customers yet!

SID177:
Yes if we add app details the issue appears. Fixing it..

kjnanda:
Closing this issue as this has been fixed and release in - V4.6.3

PR #1634, opened 2020-04-27
Fix GH-1633: Add script tag in allowed html for social sync addon

SID177:
Add script tag in allowed HTML, because social sync addon uses it.
Fixes https://github.com/rtMediaWP/rtMedia/issues/1633

### DIFF
--- a/app/main/RTMedia.php
+++ b/app/main/RTMedia.php
@@ -1748,10 +1748,10 @@ class RTMedia {
 
 		// Script, for social sync plugin.
 		$new_allowed['script'] = array(
-			'type'  => true,
-			'class' => true,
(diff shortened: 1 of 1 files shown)


app/main/RTMedia.php
```diff
@@ -1746,6 +1746,14 @@ public static function expanded_allowed_tags() {
 			'id'    => array(),
 		);
 
+		// Script, for social sync plugin.
+		$new_allowed['script'] = array(
+			'type'  => true,
```
ravatparmar, line 1751: For information: don't we need to pass array()?

SID177: Used array instead of `true`

PR #1650, opened 2020-05-18
Update issue templates

kjnanda:


pooja-muchandikar:
The changes in this PR are not in plugin instead its for repo. Repo Issue reporting template is been updated

Issue #1656, opened 2020-06-19
Issue with WP_Offload_Media when using S3 signed URL


ExoGeek:
I use wp_offload_media to offload medias to S3. Recently, I've enabled the signed URL feature (on Amazon S3). Everything goes well on the wordpress side (In the dashboard, I can see media image in WP Media library of wordpress). 

But when I visit the website (Profile -> Media), the image is not shown.

The reason is because the **URL of the media is not signed** (it links to S3 but not the signed URL).

ExoGeek:
Hello,

I've found the problem : see line 751 in the file rtmedia-function.php

`$src = apply_filters( 'rtmedia_media_thumb', $src, $media_object->id, $media_object->media_type );

// Added timestamp because conflict with cache image.
$src = $src . '?' . time();`

Maybe that you should invert the code like that : 

`// Added timestamp because conflict with cache image.
$src = $src . '?' . time();

$src = apply_filters( 'rtmedia_media_thumb', $src, $media_object->id, $media_object->media_type );`

Regards,

ExoGeek:
I've found one other place where you add "?time()" at the end of the src image (lightbox). This break Signed URL.

File : rtmedia-functions.php
Line : 591
`$html = "<img src='" . set_url_scheme( $src[0] . '?' . time() ) . "' alt='" . esc_attr( $rtmedia_media->post_name ) . "' />";`

pooja-muchandikar:
Hi @ExoGeek 

We will look into this issue and get back to you. 

Thanks and Regards,
Pooja

SID177:
Hi @ExoGeek 

I think S3 Pre-Signed URLs have query string appended, and we're adding `?` directly to the URL, which could be breaking the image.

Can you please confirm by providing image URL of any broken image. You might need to inspect element and check from browser. Or you can provide us access to 1 of your page where the images are broken.

Thanks

ExoGeek:
Hi @SID177 

You're right. rtmedia adds "?1592838041" ('?' . time();) at the end of the src image attributes.

For privacy reason, these images are not accessible publicly but you will understand the pattern.

[...]

PR #1664, opened 2020-06-25
Fix GH-1656: Check whether a URL has querystring, based on that add timestamp at the end of URL

SID177:
Check whether a URL has querystring, based on that add timestamp at the end of URL

Fixes https://github.com/rtCamp/rtMedia/issues/1656

### DIFF
--- a/app/main/controllers/template/rtmedia-functions.php
+++ b/app/main/controllers/template/rtmedia-functions.php
@@ -578,18 +578,26 @@ function rtmedia_media( $size_flag = true, $echo = true, $media_size = 'rt_media
 	if ( isset( $rtmedia_media->media_type ) ) {
 
 		if ( 'photo' === $rtmedia_media->media_type ) {
-
 			$src = wp_get_attachment_image_src( $rtmedia_media->media_id, $media_size );
 
-			/**
-			 * Used `set_url_scheme` because `esc_url` breaks the image if there is special characters are there into image name.
-			 * Added by checking the code from "wp-admin/includes/media.php:2740".
-			 * Because in media library, it was not breaking.
-			 *
-			 * Add timestamp to resolve conflict with cache image.
-			 */
-			$html = "<img src='" . set_url_scheme( $src[0] . '?' . time() ) . "' alt='" . esc_attr( $rtmedia_media->post_name ) . "' />";
+			if ( ! empty( $src[0] ) ) {
+				$src = $src[0];
 
+				// Add timestamp to bypass cache and load fresh image.
+				$src_query = wp_parse_url( $src, PHP_URL_QUERY );
+				if ( empty( $src_query ) ) {
+					$src .= '?' . time();
+				} else {
+					$src .= '&' . time();
+				}
+
+				/**
+				 * Used `set_url_scheme` because `esc_url` breaks the image if there is special characters are there into image name.
+				 * Added by checking the code from "wp-admin/includes/media.php:2740".
+				 * Because in media library, it was not breaking.
+				 */
+				$html = "<img src='" . set_url_scheme( $src ) . "' alt='" . esc_attr( $rtmedia_media->post_name ) . "' />";
+			}
 		} elseif ( 'video' === $rtmedia_media->media_type ) {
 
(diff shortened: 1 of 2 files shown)


SID177:
Yes it'll work. Any URL which has query parameter will work now. I tested it.

SID177:
Earlier the URL was becoming like this
```
https://MyS3BucketName.s3.eu-west-3.amazonaws.com/localhost/2020/06/22155525/3f083c89-a5c7-492c-938d-467d2d666dc5-1080x675.jpg?X-Amz-Content-Sha256=UNSIGNED-PAYLOAD&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=AZIA46ZC73DSWOHCZ53N%2F20200622%2Feu-west-3%2Fs3%2Faws4_request&X-Amz-Date=20200622T145545Z&X-Amz-SignedHeaders=host&X-Amz-Expires=900&X-Amz-Signature=fd92ec9003d6c72983145132cbab7df7506dec923111b2a482e045fb724140773?1592838041 
```

**?1592838041** is added. This break the URL.

The problem here is that, in `Signature` parameter, the expected value is `fd92ec9003d6c72983145132cbab7df7506dec923111b2a482e045fb724140773` but due to our code, it becomes `fd92ec9003d6c72983145132cbab7df7506dec923111b2a482e045fb724140773?1592838041`

But now it becomes like this
```
https://MyS3BucketName.s3.eu-west-3.amazonaws.com/localhost/2020/06/22155525/3f083c89-a5c7-492c-938d-467d2d666dc5-1080x675.jpg?X-Amz-Content-Sha256=UNSIGNED-PAYLOAD&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=AZIA46ZC73DSWOHCZ53N%2F20200622%2Feu-west-3%2Fs3%2Faws4_request&X-Amz-Date=20200622T145545Z&X-Amz-SignedHeaders=host&X-Amz-Expires=900&X-Amz-Signature=fd92ec9003d6c72983145132cbab7df7506dec923111b2a482e045fb724140773&1592838041
```

Here the value of `Signature` will remain `fd92ec9003d6c72983145132cbab7df7506dec923111b2a482e045fb724140773`

@umeshnevase

app/main/controllers/template/rtmedia-functions.php
```diff
@@ -578,18 +578,26 @@ function rtmedia_media( $size_flag = true, $echo = true, $media_size = 'rt_media
...
 
-			/**
-			 * Used `set_url_scheme` because `esc_url` breaks the image if there is special characters are there into image name.
-			 * Added by checking the code from "wp-admin/includes/media.php:2740".
-			 * Because in media library, it was not breaking.
-			 *
-			 * Add timestamp to resolve conflict with cache image.
-			 */
-			$html = "<img src='" . set_url_scheme( $src[0] . '?' . time() ) . "' alt='" . esc_attr( $rtmedia_media->post_name ) . "' />";
+			if ( ! empty( $src[0] ) ) {
```
umeshnevase, line 583: @SID177 Please check ! empty( $src) else it will notice that array index

SID177: @umeshnevase It won't throw any notice, `empty` will handle checking `$src` as well as `$src[0]`.
You can try
```
$a = null;
echo !empty($a[0])?'yes':'no';
```
This code in online PHP interpreter and it'll output `no`.

ExoGeek:
Hi,

Just to let you know that the suggested correction won't work for signed URL (on S3).

As soon as you append a string to the URL, the URL is not valid.

I've just tested with a valid S3 Signed URL. I've append "&1592838041" at the end of the URL and got the error below : 

```
<Code>SignatureDoesNotMatch</Code>
<Message>
The request signature we calculated does not match the signature you provided. Check your key and signing method.
</Message>
```

I suppose the best solution could be to not touch the URL if there's already a querystring.

Regards

SID177:
Hi there,

We've removed the timestamp from URL if its from `amazonaws.com`

PR #1666, opened 2020-07-03
Fix GH-1656: Add separate function to append timestamp, don't add timestamp if url is from amazon

SID177:
- Add separate function to append timestamp in URL
- Don't add timestamp if URL is from amazon AWS.

Fixes https://github.com/rtCamp/rtMedia/issues/1656

### DIFF
--- a/app/main/controllers/template/rtmedia-functions.php
+++ b/app/main/controllers/template/rtmedia-functions.php
@@ -581,15 +581,7 @@ function rtmedia_media( $size_flag = true, $echo = true, $media_size = 'rt_media
 			$src = wp_get_attachment_image_src( $rtmedia_media->media_id, $media_size );
 
 			if ( ! empty( $src[0] ) ) {
-				$src = $src[0];
-
-				// Add timestamp to bypass cache and load fresh image.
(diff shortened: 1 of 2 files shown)


app/main/controllers/template/rtmedia-functions.php
```diff
@@ -771,6 +756,24 @@ function rtmedia_image( $size = 'rt_media_thumbnail', $id = false, $echo = true,
...
+ *
+ * @param string $src Image URL.
+ *
+ * @return string Modified URL.
+ */
+function rtmedia_append_timestamp_in_url( $src ) {
+	$src_query = wp_parse_url( $src, PHP_URL_QUERY );
+	if ( empty( $src_query ) ) {
+		$src .= '?' . time();
+	} elseif ( false === strpos( $src, 'amazonaws.com' ) ) {
```
ravatparmar, line 770: Just for information: Why we've skipped amazonaws.com

SID177: https://github.com/rtCamp/rtMedia/pull/1664#issuecomment-650266868

Issue #1647, opened 2020-05-05
[Warning and Notice] Fix Warnings and Notices 


kjnanda:
- [ ] **1. When the user post the update with media on activity page, it shows below notices in the console** 

`This is present in Verison - 4.6.2 and 4.6.3 both`

![debug_log upload from activity](https://user-images.githubusercontent.com/15173772/81035892-7cf95e00-8eba-11ea-87da-246d772d7532.jpg)

- [ ] **2. When the user comments on uploaded media from the lightbox, it shows below notice and warnings**

`This is present in Verison -  4.6.3`

![debug_log](https://user-images.githubusercontent.com/15173772/81035993-f3965b80-8eba-11ea-9187-e855fcb33b15.jpg)

`2576339 ` `2578864` `2586313`

pooja-muchandikar:
@SID177 

The above warnings and notices are not shown but On uploading audio file following warning is shown


```
PHP Notice:  Trying to access array offset on value of type null in /var/www/htdocs/wp-content/plugins/rtMedia/app/main/controllers/template/rtmedia-functions.php on line 3163
```

Please check on this.

pooja-muchandikar:
Tested with `pre-release` branch and the warning/error mentioned in the comment [here](https://github.com/rtCamp/rtMedia/issues/1647#issuecomment-653025519)  are fixed.

pooja-muchandikar:
Tested with `develop` branch and the fix works as expected.

pooja-muchandikar:
Tested with `master` branch and the fix works as expected.

pooja-muchandikar:
Released with`v4.6.4`

PR #1669, opened 2020-07-10
Fix GH-1647: Add empty check before accessing values, fix phpcs

SID177:
Add empty check before accessing values. Fix PHPCS issues in function

Fixes https://github.com/rtCamp/rtMedia/issues/1647#issuecomment-653025519

### DIFF
--- a/app/main/controllers/template/rtmedia-functions.php
+++ b/app/main/controllers/template/rtmedia-functions.php
@@ -3141,7 +3141,7 @@ if ( ! function_exists( 'rtmedia_who_like_html' ) ) {
  *
  * @param object $media_object Media details object.
  *
- * @return bool|string Uploaded thumbnail URL | False.
+ * @return string|bool Uploaded thumbnail URL | False.
  */
 function rtm_get_music_cover_art( $media_object ) {
 
--- a/app/main/controllers/template/rtmedia-functions.php
+++ b/app/main/controllers/template/rtmedia-functions.php
@@ -3137,22 +3137,22 @@ if ( ! function_exists( 'rtmedia_who_like_html' ) ) {
 }
 
 /**
- * Get music cover art
+ * Get music cover art.
  *
  * @param object $media_object Media details object.
  *
- * @return      bool|string
+ * @return bool|string Uploaded thumbnail URL | False.
  */
 function rtm_get_music_cover_art( $media_object ) {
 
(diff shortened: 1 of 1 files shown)


app/main/controllers/template/rtmedia-functions.php
```diff
@@ -3163,25 +3163,25 @@ function rtm_get_music_cover_art( $media_object ) {
 	$media_tags = new RTMediaTags( $file );
 	$title_info = $media_tags->title;
 	$image_info = $media_tags->image;
-	$image_mime = $image_info['mime'];
-	$mime       = explode( '/', $image_mime );
 	$id         = $media_object->id;
 
-	if ( ! empty( $image_info['data'] ) ) {
+	if ( ! empty( $image_info['data'] ) && ! empty( $image_info['mime'] ) ) {
```
ravatparmar, line 3168: Ca we do the early return

app/main/controllers/template/rtmedia-functions.php
```diff
@@ -3163,25 +3163,25 @@ function rtm_get_music_cover_art( $media_object ) {
...
-		if ( is_array( $thumb_upload_info ) && ! empty( $thumb_upload_info['url'] ) ) {
-			$media_obj->model->update(
-				array(
-					'cover_art' => $thumb_upload_info['url'],
-				),
-				array(
-					'id' => $id,
-				)
-			);
+		if ( ! empty( $mime ) && is_array( $mime ) ) {
```
ravatparmar, line 3171: Here as well, can we do the early return

SID177: I don't think so, there's code after this condition

app/main/controllers/template/rtmedia-functions.php
```diff
@@ -3163,25 +3163,25 @@ function rtm_get_music_cover_art( $media_object ) {
...
+						'cover_art' => $thumb_upload_info['url'],
+					),
+					array(
+						'id' => $id,
+					)
+				);
 
-			return $thumb_upload_info['url'];
+				return $thumb_upload_info['url'];
+			}
```
umeshnevase, line 3184: Else return false

SID177: There's a code which is getting executed after this block. So I think returning false in else isn't a good idea.

PR #1682, opened 2020-08-04
Add rtmedia_update in editable activity types and reload page after rtmedia_update activity is saved

SID177:
- Add `rtmedia_update` type of activity in editable activity types
- Reload page when `rtmedia_update` type of activity is edited and saved.

Fixes https://github.com/rtCamp/rtMedia/issues/1681

### DIFF
--- a/app/main/controllers/activity/RTMediaBuddyPressActivity.php
+++ b/app/main/controllers/activity/RTMediaBuddyPressActivity.php
@@ -34,6 +34,8 @@ class RTMediaBuddyPressActivity {
 				// Code to show media with read more option.
 				add_filter( 'bp_activity_truncate_entry', array( $this, 'bp_activity_truncate_entry' ), 10, 3 );
 			}
+
+			add_filter( 'b_e_a_plugin_option_editable_types', array( $this, 'b_e_a_plugin_option_editable_types' ) );
 		}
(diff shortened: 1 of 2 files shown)


app/main/controllers/activity/RTMediaBuddyPressActivity.php
```diff
@@ -76,6 +78,25 @@ public function __construct() {
...
 	}
 
+	/**
+	 * Adds rtmedia_update in editable activity types when BuddyPress Edit Activity plugin is active.
+	 *
+	 * @param array $option Editable activity types.
+	 *
+	 * @return array Modified editable activity types.
+	 */
+	public function b_e_a_plugin_option_editable_types( $option ) {
```
umeshnevase, line 88: @SID177 Can you change the name of the function, keep as same as used in rtMedia plugin

SID177: I've kept the function name same as filter name, will that work?

SID177: At many places we use same function name as filter name that's why.

umeshnevase: Can you prefix with rtmedia. don't keep it the same as filter name.

SID177: Changed function name.

Issue #1486, opened 2019-11-27
[Dashbaord] Hire Us tab - Correct the back ground color of the button


kjnanda:
It should be `#007cba` color.

![Hire_Us_‹_test_—_WordPress](https://user-images.githubusercontent.com/15173772/69696206-fd928200-1104-11ea-904b-f710c3963668.png)

vaishaliagola27:
@pooja-muchandikar 

Please let me know if below button will be okay. 

<img width="625" alt="Screenshot 2020-09-10 at 5 24 35 PM" src="https://user-images.githubusercontent.com/11362577/92725950-8162e580-f38a-11ea-876f-80220649428b.png">

pooja-muchandikar:
@vaishaliagola27 

Yes, looks fine :+1:

pooja-muchandikar:
Tested with `pre-release` branch and the button looks fine. :heavy_check_mark:

pooja-muchandikar:
Tested with `master` branch and the issue is fixed.

pooja-muchandikar:
Released with `v4.6.5`

PR #1687, opened 2020-09-10
Change Button style on Hire Us Tab

vaishaliagola27:
Related to #1486


abhijitrakas (approved):
@vaishaliagola27 - Looks good to me. Only one thing, it would be better to add a screenshot in the description.

Issue #1487, opened 2019-11-27
Revisit the Validation condition on Load More button on activity page


kjnanda:
**Let me explain to you the whole scenario**

When rtmedia is enabled, the `load more` button appears on the BuddyPress activity wall. 

This button should be present only in case when the multiple activities are present. So clicking on that button it can load the hidden activities.

As per the current scenario, it is present even when there is not a single activity present on the wall! Which is really not required. Since clicking on that button will anyways show this message - `Sorry, there was no activity found. Please try a different filter.`

Please revisit the condition where the validation needs to be updated for some min number of activities. 

Please check this visual - https://drive.google.com/file/d/1p1-Dmlepe0ibKIz-uTTuS81JzWACHJGa/view

pooja-muchandikar:
@vaishaliagola27 

```
Test Environment
WordPress: v5.5.1
rtMedia: v4.6.4 pre-release branch
BuddyPress: v6.3.0 legacy and nouveau template both 
```

Cross-checked the issue with `pre-release` branch and found that the issue is fixed but another regression bug is found. Please check the following points:

- As per the issue description when there is no post in the activity page, the error message should also not be shown. 

- When there is only one activity then load more is not shown this issue is fixed, but this caused another issue. Load more is not shown even after 20 posts, instead the first post added is disappearing. 

Steps to reproduce:
- Added first activity as `Activity post 1` and later post few other posts till post 10 - **load more was not shown but the first post was visible**
- Added some more post till post 20 **load more was not shown and first post was visible**
- Now, added another post post 21 **load more was not shown and also the first post Activity Post 1 also disappeared**

After 20 posts there should be load more option and click on it should load older posts. But in this case the posts are getting disappeared. 

Refer the visual: https://drive.google.com/file/d/1Vfv92Z0LYwGJml2JKbEYmTtfbr9hM89r/view

**Note: checked for both profile and group activity page**

Pradeep1308:
```
Test Environment
WordPress: 5.8.1
rtMedia: v4.6.8 pre-release branch
BuddyPress: v9.1.1 legacy and nouveau template both 
```
This issue is not reproducible. :+1:

PR #1688, opened 2020-09-10
Remove Load More button on activity page for less Activities

vaishaliagola27:
This PR will remove `Load More` button on the activity page which was enabled even if only 1 activity was available. 

Related #1487

### DIFF
--- a/app/main/controllers/privacy/RTMediaPrivacy.php
+++ b/app/main/controllers/privacy/RTMediaPrivacy.php
@@ -47,9 +47,6 @@ class RTMediaPrivacy {
 			// show change privacy option in activity meta.
 			add_action( 'bp_activity_entry_meta', array( $this, 'update_activity_privacy_option' ) );
 
-			// To show load more button all the time.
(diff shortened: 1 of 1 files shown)


app/main/controllers/privacy/RTMediaPrivacy.php
```diff
@@ -47,9 +47,6 @@ public function __construct( $flag = true ) {
 			// show change privacy option in activity meta.
 			add_action( 'bp_activity_entry_meta', array( $this, 'update_activity_privacy_option' ) );
 
-			// To show load more button all the time.
-			add_filter( 'bp_activity_has_more_items', array( $this, 'rtm_bp_activity_has_more_items' ) );
```
abhijitrakas, line 51: @vaishaliagola27 - Do you know why we added this previously?

vaishaliagola27: @abhijitrakas No, It was there way more before.

abhijitrakas (approved):

Issue #1695, opened 2020-10-12
Unable to submit ticket through the support tab from settings


pooja-muchandikar:
Steps to Reproduce:

- Go to rtMedia settings
- Go to Support tab
- Enter all details in the form and upload attachment and click on submit button
- Observe that even after uploading attachment error is shown to upload attachment

Need to fix this issue and also if attachment is mandatory we can make it non-mandatory. 

Refer Visual: https://drive.google.com/file/d/1TmTcuqnJ1pXiJkBlbRbTsSk5FQ6d0bED/view

pooja-muchandikar:
Tested with `pre-release` branch and the issue is fixed.

pooja-muchandikar:
Tested with `master` branch and the issue is fixed.

pooja-muchandikar:
Released with `v4.6.5`

PR #1696, opened 2020-10-12
Issue with support file upload Resolve

vaishaliagola27:
This resolves issue in uploading images while support request. 

Related #1695


umeshnevase:
@pooja-muchandikar @vaishaliagola27 

 Attachment is uploading now. but We can make the attachment non-mandetary.

umeshnevase (approved):

Issue #1704, opened 2020-11-05
Sitewide activity post media is missing on /members/username/media (bp nouveau template)


razor2016:
Hi rtCamp,

Thanks for this amazing plugin.

I'm reporting a bug when using rtmedia with bp nouveau template pack.

When a user post media in the sitewide activity, it should appear on his/her profile media tab (example.com/members/username/media). This is the expected behavior and works fine with the bp legacy template.

**The problem**:
The problem only seems to occur with the bp nouveau template, and when the user uploads media from the sitewide activity, not from profile > activity (works fine there).

It's tricky, so I will try my best to explain "why and how to recreate it on your end."

**Scenario-1 (when it works):**
If the user doesn't have `#whats-new-post-in` select-box in his/her sitewide activity uploader and if he/she post a photo/video there, it appears on his/her profile media tab (members/username/media). This is expected.

![Capture](https://user-images.githubusercontent.com/20941700/98223989-180fe500-1f7d-11eb-827b-24de39ba949e.PNG)
![Capture1](https://user-images.githubusercontent.com/20941700/98223993-19411200-1f7d-11eb-93c2-c31fa4ab00b0.PNG)

**Scenario-2 (when it doesn't work):**
Let's say the user has joined a group, so now he/she has `#whats-new-post-in` select-box to select from, when posting a photo or video. Now if the user post a photo in the sitewide activity, that photo doesn't appear in his/her profile (members/username/media).

![Capture2](https://user-images.githubusercontent.com/20941700/98225274-bbadc500-1f7e-11eb-989a-08cd7fc123f4.PNG)
![Capture4](https://user-images.githubusercontent.com/20941700/98225279-bcdef200-1f7e-11eb-9fce-036b39839c87.PNG)

**The cause of the problem**: 
I tried to find the cause of the problem and what I found is:
In Scenario-1, the context_id of the uploaded media has a value 1 (see the `wp_rt_rtm_media` db table).
in Scenario-2, the context_id of the uploaded media has a value 0/null (see the `wp_rt_rtm_media` db table).

I didn't go further, checking why it gets 0/null in the scene-2. I am leaving that too you since everyone will benefit from it.

To recreate the issue:
1. you must use the bp nouveau template pack.
2. The user must have `#whats-new-post-in` select-box in his/her activity uploader. Perhaps join a group so that the `#whats-new-post-in` select-box would appear.
3. user has to post media at the sitewide activity (/activity), not at member/username/activity.

[...]

PR #1705, opened 2020-11-12
Fix GH-1704: Add support to post activity with media from sitewide activity stream from Nouveau template

SID177:
Sets proper context_id and context when posting activity with media from sitewide activity stream in Nouveau template
Also allows to select a particular group to post media and activity to

Fixes https://github.com/rtCamp/rtMedia/issues/1704


abhijitrakas (approved):
LGTM 👍

PR #1713, opened 2020-12-17
Add try catch and undefined check before checking action of ajax call

SID177:
- Add try catch and undefined check before checking action of ajax call
- To fix the conflict with EventON plugin, because for some ajax calls it doesn't pass `action` parameter


pooja-muchandikar:
Tested `fix/conflict-eventon-plugin` branch and the fix added is not conflicting with edit activity.

vaishaliagola27 (approved):

pooja-muchandikar:
Tested with `pre-release` branch and the fix added is not conflicting with edit activity.

pooja-muchandikar:
Tested with `develop` branch and the fix added is not conflicting with edit activity.

pooja-muchandikar:
Released with `v4.6.6`

Issue #1723, opened 2021-01-05
Javascript string translation errors


pooja-muchandikar:
Need to work on following ticket, details are mentioned in the ticket. https://wordpress.org/support/topic/javascript-string-translation-errors/#post-13865199

NicestRudeGuy:
@vaishaliagola27  
Inside **rtMedia plugin** we have used wp_localize_script method several places 
for example https://github.com/rtCamp/rtMedia/blob/develop/app/main/RTMedia.php#L1319
`
wp_localize_script( 'rtmedia-main', 'rtmedia_gallery_reload_on_upload', '1' );
`
but inside the official documentation the last param is supposed to be an array https://developer.wordpress.org/reference/functions/wp_localize_script/

the issue arrises when we the person uses php v8 with WP. currently wordpress 5.6 supports php 7.4 but php 8.0 support is in beta so in near future we will face this issue ( warnings ) . several of our addons might be affected if we change it.

Also for single data do we need to pass array or we can just pass the data directly ?

vaishaliagola27:
@NicestRudeGuy 

We have use single dimensional array as value and change in respective js files.

NicestRudeGuy:
@vaishaliagola27 
These only exist in PHP 8.0 and are everywhere even in the addons will require update in all the addons as well .  Cant seem to find a better way to solve this problem except manually updating rtMedia plugin and all the addons

SID177:
Maybe we can do something like this:
- Make changes in core rtMedia plugin and test it, keep the old code for now so it works with older versions of premium add-ons and themes
- By referring to these changes, make changes in premium add-ons and themes, test it and release them, make sure to add backward compatibility in add-ons and themes so that it works with recent rtMedia versions
- Release rtMedia core with the changes
- After 2-3 releases, maybe we can remove the older code for `wp_localize_script` from core plugin and backward compatibility code from add-ons and themes

NicestRudeGuy:
> Maybe we can do something like this:
> - Make changes in core rtMedia plugin and test it, keep the old code for now so it works with older versions of premium add-ons and themes
> - By referring to these changes, make changes in premium add-ons and themes, test it and release them, make sure to add backward compatibility in add-ons and themes so that it works with recent rtMedia versions
> - Release rtMedia core with the changes
> - After 2-3 releases, maybe we can remove the older code for `wp_localize_script` from core plugin and backward compatibility code from add-ons and themes


[...]

PR #1724, opened 2021-01-07
Fix/issue GH-1723

NicestRudeGuy:
### Introduction 
When using **PHP v8** with wordpress there are several notices regarding localization function , it requires you to pass a array in the 3rd param , in the codebase we have used single var which doesn't conflicts but it started giving warnings/notices in admin dashboard as well as every page.

_REF_: https://wordpress.org/support/topic/javascript-string-translation-errors/#post-13865199

### Solution
I grouped similar sort of variables where possible and single arrays inside conditional statements . Added conversion in the top of the **JS** files so that we wont have to go in the code and change every variable manually. 

### Tested on PHP 7 and 8 no issues found


VijayaRaghavanM:
Closing this PR as the issues are already fixed in https://github.com/rtCamp/rtMedia/pull/1728

Issue #1727, opened 2021-03-08
Hide the download option for video player


pooja-muchandikar:
There is option to download the video file even if there is not download plugin enabled. This option is default player option which needs to be hidden.

## Solution 

Fix is to add a custom code/make code changes to hide this.

pooja-muchandikar:
https://github.com/rtCamp/rtMedia/pull/1743#issuecomment-919876863

pooja-muchandikar:
Cross-checked this issue with `develop` branch and the fix looks fine. 
<img width="951" alt="Screenshot 2021-09-16 at 1 13 52 AM" src="https://user-images.githubusercontent.com/41000648/133499124-a9c9deca-8f32-40bd-98dc-457bf5fd3997.png">

pooja-muchandikar:
Cross-checked the issue with `master` branch and works as expected.

pooja-muchandikar:
Issue fixed and released in version `4.6.7`

<img width="1071" alt="Screenshot 2021-09-16 at 7 44 00 PM" src="https://user-images.githubusercontent.com/41000648/133628082-6cc22881-eece-4907-a1a6-e9166df65b2e.png">

PR #1734, opened 2021-06-24
Fix GH-1727: hide download option for video player

kapilpaul:
This PR will hide the download option from video player.

issue: #1727


OpuRockey (approved):

pooja-muchandikar:
@kapilpaul 
if you are available can you please check?
I can still see the download option

image

PR #1744, opened 2021-09-16
fix: manual upload with/without text

kapilpaul:
fix: manual upload with/without text

### DIFF
--- a/app/assets/js/rtMedia.backbone.js
+++ b/app/assets/js/rtMedia.backbone.js
@@ -1186,6 +1186,18 @@ jQuery( document ).ready( function( $ ) {
 		}
 	} );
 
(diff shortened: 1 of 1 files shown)


app/assets/js/rtMedia.backbone.js
```diff
@@ -1186,6 +1186,18 @@ jQuery( document ).ready( function( $ ) {
 		}
 	} );
 
+	// change color of what's new if content is &nbsp;
+	let whatsNew = jQuery( '#whats-new' );
```
SID177, line 1190: It seems this var is not being changed anywhere, use `const` instead

SID177 (approved):

Issue #1548, opened 2020-02-06
Count on Media tab remains on adding/deleting media


pooja-muchandikar:
```
Test Environment
WordPress: v5.3.2
rtMedia: v4.5.11
BuddyPress: v5.1.2
```

**Steps to reproduce:**
1. Go to Media page
2. Upload media 
3. Check the count of media on media tab. The count doesn't increase. 
ex: if the previous count is 4 then on uploading new image the count should be 5 but the count is not shown. The count is shown but on explicitly reloading the page which shouldn't be. Count should be shown immediately. 

**Note: Same thing applies while deleting media as well.**

Refer the visual -> https://drive.google.com/file/d/1FcgdavcB05zxWtNWii1IB8q2dqlRjoAD/view

vermadarsh:
@pooja-muchandikar 
Added 1 filter and 2 actions in this fix.

Filter
`rtmedia_media_delete_success_message` - This filter helps in me modifying the success message that is printed after media is deleted successfully.

Actions
`rtmedia_before_media_delete` - This action fires just before the media is deleted.
`rtmedia_after_media_delete` - This action fires just after the media is deleted.

pooja-muchandikar:
https://github.com/rtCamp/rtMedia/pull/1751#issuecomment-933410549 

https://github.com/rtCamp/rtMedia/pull/1752#issuecomment-933410069

PR #1752, opened 2021-10-04
fix: count media on upload

kapilpaul:
issue: #1548


pooja-muchandikar:
@kapilpaul 

Cross-checked this PR and the fix is working as expected. Please merge the PR.

Issue #1010, opened 2017-11-02
Responsive devices --> Portrait mode - Gallery media search box alignment issue on


kjnanda:
Please check this screenshot.
1. iPhone 6S 
![unnamed image 6](https://user-images.githubusercontent.com/15173772/32319634-4477ad16-bfe1-11e7-98ae-f42354ba6057.png)

2. Samsung Galaxy S8
![samsung galaxy s8](https://user-images.githubusercontent.com/15173772/32319948-2e456a78-bfe2-11e7-8e60-6aaec66a9959.png)

PR #1763, opened 2021-10-18
fix media search box

Pradeep1308:
**What Fixed? :hammer:** 
Gallery media `search box` alignment

ref: #1010


app/assets/admin/css/admin.css
```diff
@@ -159,8 +159,8 @@
   -webkit-box-flex: 0;
       -ms-flex: 0 0 auto;
           flex: 0 0 auto;
-  padding-right: 0.5rem;
-  padding-left: 0.5rem;
+  padding-right: .5rem;
```
kapilpaul, line 162: please do not remove the leading 0. fix on the other lines too where you removed.

kapilpaul: Grunt build files are .min.css . please check again if its converted by grunt.

kapilpaul (changes requested):

Issue #1296, opened 2019-03-06
Issue list with RTL(right-to-left) pattern


kjnanda:
```
Test Environment 

1. Wordpress 5.1
2. rtMedia Core 4.5.6
3. BuddyPress - 4.2.0
4. RTL Tester Plugin 1.1 (https://wordpress.org/plugins/rtl-tester/)
```
---------

## 1. Backend 

Please correct some minor changes which appears incase of RTL.

- [x] 1.1 `rt Logo` need some space. 

![logo](https://user-images.githubusercontent.com/15173772/53867806-9a167080-401a-11e9-85dd-1c38eb164d68.png)

------

- [x] 1.2 Font family gets changed for the mentioned class below. Please check this under `rtMedia Settings Screen`.

```
1. .rtm-tabs
2. .rtm-option-title
3. .hndle
4. .nav-tab-wrapper
5. .notice.notice-info.install-transcoder-notice.is-dismissible 
6. .updated.rtmedia-inspire-book-notice
7. .updated.rtmedia-pro-split-notice
```

Please check this link - https://nimb.ws/Cghtyb

### Expected Font Family type

```
system-ui - 700×
Family
-apple-system, system-ui, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif;
```

### Actual Font Family type 
 
```
Tahoma - 700×
Family
Tahoma, Arial, sans-serif;
```
-------------

## 2. Frontend

- [x] 2.1  Album Page - `Pipe` symbol is misaligned. It should be between `Upload and Option button`
https://goo.gl/471KU4

- [x] 2.2 Listed below items should be right aligned.

```
1. Create an Album label
2. Album Title Label
3. Album Description Label
4. Create Album Button
5. The message is shown after the album is created
```
Please check this screen - https://goo.gl/BMc1o1

- [x] 2.3 Clicking upon `Option` button - shows `Add Album` button which goes out of boundary for some themes. 

1. Theme 2012 - https://goo.gl/ZqKZFu
2. Theme 2016 - https://goo.gl/eL4TtL

- [x] 2.4 Opening media in lightbox - Please check below details which are not aligned RTL pattern.

pavanpatil1:
Cross-checked the issue. The fix is working fine. 

> 1.1 rt Logo needs some space.

The logo is shifted. 

[...]

PR #1786, opened 2021-11-18
Fix/rtl

Pradeep1308:
**What Changed? :hammer:** 

Fix issues listed in ref: #1296 according to RTL pattern.


app/admin/RTMediaAdmin.php
```diff
@@ -1229,13 +1229,14 @@ public static function get_current_tab() {
...
-			?>
 
+			 $align = is_rtl() ? 'alignleft' : 'alignright';
+			?>
 			<div class="wrap bp-media-admin <?php echo esc_attr( $this->get_current_tab() ); ?>">
 				<div id="icon-buddypress-media" class="icon32"><br></div>
 				<div>
 					<h2 class="nav-tab-wrapper"><?php $this->rtmedia_tabs(); ?>
-						<span class="alignright by">
+						<span class="<?php printf( esc_html( $align ) ); ?> by">
```
sabbir1991, line 1239: Here need to use `esc_attr` not `esc_html` like this
We can also write this way `<?php echo esc_attr( $align ); ?>` instead of `<?php printf( esc_html( $align ) ); ?>`

Pradeep1308: On it 👍

sabbir1991 (approved):

Issue #1645, opened 2020-05-04
Allow the subscriber to Modify Image functionality 


kjnanda:
![Media_–_Krupa_Nanda_–_krupa_dev5_rt_gw](https://user-images.githubusercontent.com/15173772/80940079-380ef200-8dfc-11ea-9630-438a79432dbd.png)

shapaaa:
After discussion with Mehedi,Pradeep & Sabbir there is some security concerns by giving this permission to subscriber & there is no use case of subscriber role as all roles will get edit_post capability.

PR #1788, opened 2021-11-29
Allow the subscriber to Modify Image functionality

shapaaa:
ref: #1645


app/main/controllers/template/rtmedia-actions.php
```diff
@@ -81,7 +81,7 @@ function rtmedia_image_editor_content( $type = 'photo' ) {
 		$media_id      = $rtmedia_query->media[0]->media_id;
 		$id            = $rtmedia_query->media[0]->id;
 		$modify_button = '';
-
+		get_role( 'subscriber' )->add_cap( 'edit_posts', true );
```
sabbir1991, line 84: I am not sure the use case here. But in my opinion, this is not a good idea to add `edit_posts` permission to subscriber role. When we add this caps in this role, it's always present on subscriber role and maybe it makes conflict with other plugin. Alternative you can check `map_meta_cap` filter for adding on demand caps. Please check this.

shapaaa: I did check  `map_meta_cap` 
But I am not sure how it will allow subscriber to edit media 
I also checked out this [blog](https://elicus.com/use-of-map-meta-cap-wordpress-filter/) where he adds custom cap in admin_init hook & uses map_meta_cap when he wants to give access to specific post.
But here as per requirement subscriber should  be able to edit any media uploaded by them
Please let me know how should I change code then.

PR #1792, opened 2021-12-09
Fix/modify image

shapaaa:
ref: #1645


shapaaa:
@sabbir1991 please check this PR I have fixed issue by changes you mentioned

sabbir1991 (approved):

shapaaa:
After discussion with Sabbir ,Pradeep & Mehedi the soln to this includes security concerns & also there will not be any use case for subscriber role as if all roles will have edit post capability.

PR #1794, opened 2021-12-15
Merge Pre-release to Develop 

Pradeep1308:
Merge Pre-release to Develop


templates/media/album-gallery-item.php
```diff
@@ -26,7 +26,9 @@
 	$rtmedia_backbone['is_edit_allowed'] = $is_edit_allowed[0];
 }
 ?>
-<li class="rtmedia-list-item" id="<?php echo esc_attr( rtmedia_id() ); ?>">
+
+<?php // rtmedia_id is return backbone object if we use esc_attr then it will create > & < into &lt; &gt; it will not interpret backbone object into media id. ?>
+<li class="rtmedia-list-item" id="<?php echo rtmedia_id(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>">
```
sabbir1991, line 31: Here we can use `esc_attr` function instead of phpcs:ignore

templates/media/album-gallery-item.php
```diff
@@ -26,7 +26,9 @@
...
 	<a href="<?php rtmedia_permalink(); ?>" title="<?php echo esc_attr( rtmedia_title() ); ?>">
```
sabbir1991, line 32: Please use `esc_url` in `rtmedia_permalink()` function

sabbir1991 (approved):

templates/media/album-gallery-item.php
Pradeep1308: rtmedia_id is return backbone object if we use esc_attr then it will create > & < into &lt; &gt; it will not interpret backbone object into media id.
i didn't do it because of this comment.

PR #1856, opened 2022-03-15
fix checks for media

Pradeep1308:
Fixes the  [issue](https://github.com/rtCamp/rtMedia-Pro/issues/186) for add-on shortcode generator

### DIFF
--- a/app/main/controllers/template/rtmedia-functions.php
+++ b/app/main/controllers/template/rtmedia-functions.php
@@ -1559,7 +1559,7 @@ function rtmedia_pagination_next_link() {
 
 			// if there are more media than number of media per page to show than $rtmedia_query->media->media_id will be set other wise take media_id of very first media.
 			// For more understanding why array became object check rewind_media() in RTMediaQuery.php file and check it's call.
-			if ( false !== $rtmedia_query->media ) {
+			if ( ! empty( $rtmedia_query->media ) ) {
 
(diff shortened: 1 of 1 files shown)


pavanpatil1:
Cross-checked the issue, Fix is working fine. You can merge the PR @Pradeep1308 once the review is done. 


https://user-images.githubusercontent.com/44057535/158581213-9cb56144-d120-4190-b658-0def83fe2f75.mp4

app/main/controllers/template/rtmedia-functions.php
```diff
@@ -1559,13 +1559,16 @@ function rtmedia_pagination_next_link() {
 
 			// if there are more media than number of media per page to show than $rtmedia_query->media->media_id will be set other wise take media_id of very first media.
 			// For more understanding why array became object check rewind_media() in RTMediaQuery.php file and check it's call.
-			$post_id = ( isset( $rtmedia_query->media->media_id ) ? $rtmedia_query->media->media_id : $rtmedia_query->media[0]->media_id );
-			$post    = get_post( get_post_field( 'post_parent', $post_id ) );
+			if ( false !== $rtmedia_query->media ) {
```
kamalahmed, line 1562: It is good to use !empty() check instead of checking strictly with a boolean. because current check will cause error if the check returns any falsy value which is not a bolean false, like empty string or null etc will cause issue.

kamalahmed (changes requested):
Please check the comment and update for better error protection. 😄

kamalahmed (approved):
looks good 👍

Issue #1890, opened 2022-04-15
Issue observed on Pre-release branch


pavanpatil1:
```
Environment: 
WordPress: v5.9.3
buddypress: v10.2.0
rtMedia: Pre-release branch 
```

- [x] PHP notice is displayed after activating the plugin. 

Below PHP notice is displayed on debug log after activating the plugin. 

`PHP Notice:  Trying to access array offset on value of type bool in  \app\public\wp-content\plugins\rtMedia\app\admin\RTMediaAdmin.php on line 110`

https://user-images.githubusercontent.com/44057535/163426556-005d1768-bafc-4752-a59a-0e07bdc4318b.mp4

---

- [ ] Count for the album is visible but the album is not visible on the Groups albums page. 

On the Groups albums page, the count for the default album is visible but the album is not visible. 

**Steps to reproduce:** 
1. Install and activate the rtMedia plugin. Create a group and enable media in the group from rtMedia -> BuddyPress. 
2. Navigate to the group media page and click on albums. 
3. Observe the issue. 

Expected: Count should not be visible if the album is not present. 

https://user-images.githubusercontent.com/44057535/163515824-023e9cf5-92e8-4fec-b74f-69b1b372f35a.mp4

---

- [x] Console error for Masonry view is displayed on the member's activity page. 

**Steps to reproduce:** 
1. Enable the Masonry view from rtMedia -> Display settings. 
2. Navigate to the Activity/Members activity page and observe the console error. (Add a comment on any media to view the console error). 

Below console error is displayed. 
`masonry not initialized. Cannot call methods, i.e. $().masonry("reload")`

![image](https://user-images.githubusercontent.com/44057535/163516578-7d54c3b1-b2e8-45e1-b9df-0f98a3ffe719.png)

---

- [x] DB error displayed while editing the album. 

While editing the album, the database error is displayed in debug log while moving the media to another album. 

[...]

PR #1891, opened 2022-04-15
GH : #1890 Issue observed on Pre-release branch

Mukulsingh27:
1. #1890 

- [x] PHP notice is displayed after activating the plugin.
- [x] Count for the album is visible but the album is not visible on the Groups albums page.
- [x] Console error for Masonry view is displayed on the member's activity page.
- [x] DB error displayed while editing the album.

### DIFF
--- a/app/main/controllers/template/RTMediaNav.php
+++ b/app/main/controllers/template/RTMediaNav.php
@@ -240,7 +240,7 @@ class RTMediaNav {
 	 * Add submenus for My account.
 	 */
 	public function sub_nav() {
-		global $rtmedia, $rtmedia_query, $wp;
+		global $rtmedia, $rtmedia_query;
 
 		$active_components = bp_get_option( 'bp-active-components' );
 		if ( isset( $active_components['groups'] ) ) {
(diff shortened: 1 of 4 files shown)


pavanpatil1:
Verified the PR. Fixes are working fine. 

>  Count for the album is visible but the album is not visible on the Groups albums page.

Just in the album count issue, the count for the default album is not visible on the groups' album page. Rest all three issues working fine. @Mukulsingh27 Could you please check this. 

countgrp

app/main/controllers/template/RTMediaNav.php
```diff
@@ -240,7 +240,7 @@ public function admin_nav() {
 	 * Add submenus for My account.
 	 */
 	public function sub_nav() {
-		global $rtmedia, $rtmedia_query;
+		global $rtmedia, $rtmedia_query, $wp;
```
sabbir1991, line 243: I think you can remove the `$wp` global variable from here. Since we are not checking anything using global `$wp`

sabbir1991 (changes requested):

app/main/controllers/template/RTMediaNav.php
Mukulsingh27: Done

sabbir1991 (approved):
LGTM 👍🏼

PR #1979, opened 2023-07-24
Multiple shortcodes cannot work on the same page

Utsav-Ladani:
# Multiple shortcodes cannot work on the same page

- `rtmedia_gallery` shortcode is not working correctly when adding it more than once.
- Reason behind this is the pagination is designed that way.
- Tried to fix this and was able to with `backbone.js`, but it works only for the `Load More` type pagination.
- `1,2,3...6` numerical pagination is based on the URL. All data is loaded using the URL `/pg/2/`. 
- In this case, two different shortcodes with different pages will not work.
- So this PR adds restriction and prevents the shortcode from generating output if It is used more than once.


Pradeep1308 (dismissed):

kjnanda:
@Pradeep1308 Can you please share the visual of the issue and fix you added? This will help us to replicate and cross check.

Issue #1934, opened 2022-06-24
Deactivation Reason: collecting user data while deactiving the plugin


thebengalboy:
**Description**
We want to collect feedback when the plugin is deactivated on the site. 

**What we will collect:**
- User email who deactivated the plugin
- Reason why deactivated
- Site URL

**Procedure:**
- We need to make an API call to rtMedia site to collect feedback.
- Display popup when user clicks on deactivate link of our plugin
- When users click on submit & deactivate, make an API call to our site to submit feedback and the plugin will be deactivated.
- We can make the API call in the plugin deactivation hook and plugin deactivation will be handled by WP itself.
- Create a custom database table that will store feedback on [rtMedia.io](http://rtmedia.io/) site and create a custom admin page to display those feedback in WP admin. It should have an export option in CSV.
- Additionally, add a yes/no question saying "Notify me about future updates and features" with neither option selected by default and let select by the user.

**Feedback Title**: If you have a moment, please let us know why you are deactivating:
**Feedback options:**
- I couldn't understand how to make it work
-  I found a better plugin
-  The plugin is great, but I need specific feature that you don't support
-  The plugin is not working
-  It's not what I was looking for
-  The plugin didn't work as expected
- Other (Other will have a text input field)

Example Prototype: https://prnt.sc/xnqHwpmctoLj

robicse11127:
@chandrapatel @gagan0123 
After doing RND to find out how we can implement this feature into our plugin. I found out most of the plugins who are using this feature are using service from [Freeminus](https://freemius.com/) or [Appsero](https://appsero.com/) or somethings similar to this. Who are provides a SDK to integrate in the plugin or theme and track down the survey data with the popup and getting the used feedback when deactivating. 
So will we you the same for us? Please let me know your thoughts.
Thanks.
Cc: @thebengalboy

thebengalboy:
Please note that we may not be able to afford a paid service for the time being. As a result, we need to seek an appropriate free choice.

thebengalboy:
@robicse11127 Is there any drawback to the method suggested by @chandrapatel if we do not use a third-party service? What are the additional benefits of having a 3rd party service?

robicse11127:
The benefit of using those services are...
- They provide SDK to integrate in the plugin which can be implemented easily and has customisable options from the code as we want.
- It has data driven dashboard which provides various reports about the plugin

And if we want to create from the scratch
- Currently we do not have API ready in the main site where that data will be submitted. 
[...]

PR #1989, opened 2023-08-07
Fix the plugin deactivation survey issue

Utsav-Ladani:
# Fix the plugin deactivation survey issue

- Change the code to handle the different responses of the AJAX call.
- Issue was with the `HTTP` authentication header.
- [rtmedia-dev.rtm.rt.gw](https://rtmedia-dev.rtm.rt.gw/) site is protected with `HTTP` authentication.
- On the production site, there is no HTTP authentication, so it will work fine.
- To test the feature, please disable the `HTTP` authentication from `Easydash` and enable it after the testing.

## Related Issue
- #1934


Pradeep1308 (approved):

kjnanda:
@Utsav-Ladani This is working as expected.

PR #2014, opened 2023-11-16
Change return value from NULL to false

Utsav-Ladani:
- Change the return value from `NULL` to `false` to fix the PHP deprecation error.


Pradeep1308 (approved):

pavanpatil1:
Hi @Utsav-Ladani , Verified the PR, it is working fine. The deprecation notice is not visible now


image

PR #2143, opened 2025-05-29
Update: GoDAM Integration [ pre-release ]

krishana7911:
## Update: GoDAM Integration


app/assets/js/godam-ajax-refresh.js
```diff
@@ -0,0 +1,304 @@
...
+
+    observer.observe(container, {
+      childList: true,
+      subtree: true
+    });
+  });
+});
+
+// Debug function to test AJAX connectivity
+function testAjaxConnection() {
```
KMchaudhary, line 278: Looks like this function is not in use, we can remove if not required

PR #2214, opened 2025-11-20
Bug Fix : Fix the Bug in .js files due to build process 

NoumaanAhamed:
The build step `npx grunt build` introduced changes in the .min.js / .min.css and other .js/.css  files which caused the CSS in the dashboard to change. 

This Bug fix essentially removes the unnecessary changes.  

image


KMchaudhary (approved):
As discussed with @NoumaanAhamed, the issue occurred because new build files were generated using the npx grunt build command. In this PR, @NoumaanAhamed reverted the JS build files to those from v4.7.4, which temporarily resolves the problem. However, this suggests the underlying issue is in the original source JS.

I recommend addressing the root cause in a separate issue for the next release.

PR #2363, opened 2026-08-17
Version update v4.7.12

the-hercules:
Version bump, changelog and translation template for the 4.7.12 release.

## Changes

| file | change |
|---|---|
| `index.php` | `Version:` header and `RTMEDIA_VERSION` constant → `4.7.12` |
| `readme.txt` | `Stable tag:` → `4.7.12`, changelog entry, Upgrade Notice entry, contributor added |
| `README.md` | changelog entry |
| `changelog.txt` | changelog entry |
| `languages/buddpress-media.pot` | regenerated |

`Requires at least: 4.9.6` and `Tested up to: 7.0` were already updated in #2357 and are unchanged here. Verified against the wp.org core API that current WordPress is 7.0.4, so `7.0` is valid.

## Changelog wording

Deliberately generic, matching the 4.7.10 / 4.7.11 style. The underlying fixes are authorization and privacy related, and the specifics are not spelled out so that unpatched installs are not handed a roadmap.

## POT

Regenerated with `wp i18n make-pot . languages/buddpress-media.pot` (note: there is no `makepot` task in the Gruntfile, so `grunt makepot` does not apply). It picked up the strings added by #2357:

- `You do not have permission to submit a support request.`
- `user registration is not allowed`
- `Upload failed due to internal server error.` (also corrects the previous `Uploade failed…` typo)

and dropped `Thank you for your time.`, left over from the removed dead `linkback` / `convert_videos` handlers. The remaining line churn is source-reference renumbering.

## Build assets

Ran the full production build (`npx grunt build`) on the `.nvmrc`-pinned node v20.18.1. It produced **no file changes** — the committed minified assets already match what the sources compile to, so no asset commit is included.

This was checked deliberately, because #2357 added AJAX nonce wiring to admin JavaScript and a rebuild would overwrite any edit that had been made only to a minified file. Every nonce marker present in `admin.min.js`, `rtmedia-admin.min.js`, `importer.min.js` and `migration.min.js` traces back to its source file, and the marker sets are identical before and after the rebuild.

## Verification

Version string is consistent across `index.php` header, `RTMEDIA_VERSION`, `readme.txt` stable tag and the POT `Project-Id-Version`.

Admin AJAX behaviour re-checked at runtime against the rebuilt assets:

| actor | nonce | result |
|---|---|---|
| subscriber | valid, own | blocked, option unchanged |
| administrator | valid | succeeds |
| administrator | invalid | blocked |

## Release date

The changelog is dated **August 17, 2026**. If the tag is pushed on a different day, this needs updating before tagging.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,10 +149,10 @@ https://www.youtube.com/watch?v=dJrykKQGDcs
 
 ### 4.7.12
 * FIXED
-  * Improved authorization checks for media, albums, comments and activity privacy so actions are limited to permitted users.
-  * Improved privacy enforcement for media returned by the JSON API.
-  * Improved validation and sanitization of upload targets, album selection and query parameters.
-  * Added capability and nonce verification to administrative AJAX actions.
+  * Fixed authorization checks for media, albums, comments and activity privacy so actions are limited to permitted users.
+  * Fixed privacy enforcement for media returned by the JSON API.
+  * Fixed insufficient validation and sanitization of upload targets, album selection and query parameters.
+  * Fixed missing capability and nonce verification on administrative AJAX actions.
   * Fixed PHP 8 warnings reported on page load.
 
 * ENHANCEMENT
--- a/README.md
+++ b/README.md
@@ -147,6 +147,18 @@ https://www.youtube.com/watch?v=dJrykKQGDcs
 
 ## Changelog ##
 
+### 4.7.12
+* FIXED
+  * Improved authorization checks for media, albums, comments and activity privacy so actions are limited to permitted users.
+  * Improved privacy enforcement for media returned by the JSON API.
+  * Improved validation and sanitization of upload targets, album selection and query parameters.
+  * Added capability and nonce verification to administrative AJAX actions.
+  * Fixed PHP 8 warnings reported on page load.
+
+* ENHANCEMENT
+  * Improved compliance with WordPress plugin coding and security standards.
+  * Removed development-only files from the distributed plugin package.
+
 ### 4.7.11
 * FIXED
   * Security enhancement: Improved data validation and sanitization for media queries and search filters.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,12 +1,12 @@
 == Changelog ==
 
-= 4.7.12 [August 17, 2026] =
+= 4.7.12 [August 18, 2026] =
 
 * FIXED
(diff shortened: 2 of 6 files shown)


the-hercules:
Thanks — addressing both review comments.

**1. Stripped `php-format` flags — valid, fixed.**

Correct catch. The cause was a wp-cli version mismatch on my side: I generated with **WP-CLI 2.10.0**, while every prior release POT (4.7.10, 4.7.11) was generated with **2.12.0**. 2.12.0 emits `#, php-format`; 2.10.0 does not.

Regenerated with 2.12.0. The flag count is back to 38, matching the previous POT, and `X-Generator` now reads `WP-CLI 2.12.0` consistently with prior releases. The diff no longer removes a single `#, php-format` line, and the file shrank from 763 to 724 deletions.

The release's actual string delta is unchanged — 5 added, 4 removed:

```
+ "You do not have permission to submit a support request."
+ "user registration is not allowed"
+ "Upload failed due to internal server error."   (corrects the old "Uploade failed…" typo)
- "Thank you for your time."                      (dead linkback/convert_videos handlers, removed in #2357)
```

(`All Group Members` / `Group Admins and Mods only` appear on both sides — same strings, moved source references.)

**2. Misspelled POT filename — accurate, but out of scope for this PR.**

The observation is correct: `languages/buddpress-media.pot` is missing a `y`, and a correctly-spelled `languages/buddypress-media.pot` also exists, stale at 4.7.10.

Not changing it here, for three reasons:

- The misspelled file is the one this repo has actually maintained for many releases (4.7.9, 4.7.11 and earlier all updated it), and it is the file named in the documented release process. Silently switching filenames during a release changes the release procedure.
- There is no user-facing impact. A `.pot` is a translator template and is never loaded at runtime — WordPress loads `.mo` files, which are correctly named `buddypress-media-{locale}.mo`. Plugin translations are also built by translate.wordpress.org from a source scan, not from the shipped POT.
- Deleting a tracked file and renaming another is a real change that deserves its own PR and review, not a rider on a version bump — particularly on a release that exists to close out a security remediation.

Worth doing as separate housekeeping: rename to the correct spelling and delete the stale duplicate. Happy to raise that issue separately.

NoumaanAhamed (approved):

